PR #1, opened 2020-09-18
Remove the need for FUSE by extracting the AppImage

jonathanperret:
For the AppImage contents to work correctly, the files contained in the AppImage must be accessible at their absolute paths (that's what the AppImage runtime does, overlaying the FUSE-mounted filesystem over the root FS).

Since we're building a Unity-specific docker image, there is no harm in "permanently" copying those files at the root of the image's filesystem. This then makes the AppImage's entrypoint `AppRun` work.

This PR also contains an example of running Unity Hub from the `Dockerfile`. Note the small wrinkle of having to wait for `Xvfb` to clean up its lockfile.


GabLeRoux:
Jonathan, you’re amazing. Thank you

PR #7, opened 2020-10-02
Revert "Tmpfs mount to circumvent xvfb lock file issues"

webbertakken:
Reverts Unity-CI/docker-linux#5

Unfortunately we found out that there are some (not-MVP-worthy) challenges with `--mount=type=tmpfs,target=/tmp` during build time, as it's an experimental feature of BuildKit. 

During runtime it could work but during build time it was not writing to the tmpfs, or not extending the volume to be bigger than a few megs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -21,9 +21,6 @@ RUN apt-get -q update \
     lsb-release \
     && apt-get clean
 
-# Inject patched binaries
-COPY xvfb-run /usr/bin/
-
 # Environment
 ENV UNITY_DIR="/opt/unity"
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -22,12 +22,14 @@ RUN apt-get -q update \
     lsb-release \
     && apt-get clean
 
+# Inject patched binaries
+COPY xvfb-run /usr/bin/
+
```
webbertakken, line 27:
```suggestion
```

Dockerfile
```diff
@@ -9,6 +8,7 @@ RUN apt-get -q update \
     libxtst6 \
     libxss1 \
     desktop-file-utils \
+    fuse \
```
webbertakken, line 11:
```suggestion
```

Leave fuse out

PR #9, opened 2020-10-02
Multi stage editor build

webbertakken:
Changes:
- Split up the Unity Hub image into a `base`  and a `hub` image
- Create initial version of editor image


editor/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+###########################
+
+FROM hub AS builder
+
+ARG version="2020.1.4f1"
+ARG changeset="fa717bb873ec"
+ARG module="android"
+
+# Install
+RUN unity-hub install --version $version --changeset $changeset --module $module
```
RiederAlex, line 12: I think '--cm' needs to be added here - this automatically installs the child modules (e.g. for android openjdk,sdk,ndk)

webbertakken: Currently testing if that works for the `install` command, as it's only documented for the `install-modules` command.

webbertakken: works

PR #19, opened 2020-10-03
build all images

webbertakken:
#### Changes

- Builds unity hub on PR, based on the base image.

#### Checklist


webbertakken:
Currently this doesn't seem to work because the build action is always forcing docker.io registry, ignoring local tags.

webbertakken:
Currently we're stuck because GitHub Actions hangs when exporting the docker image.

When retrieving the logs, the file for the build editor step is not even there and the view raw logs also does not contain anything. So we're stuck for now. :(

 I have created a support ticket with GitHub.

BLaZeKiLL:
I see the actions are using `buildx` (aka buildkit) tough we are not using and buildkit specific feature did the normal build command also throw the same error ?

webbertakken:
@BLaZeKiLL Buildkit is becoming the new de facto way to build it seems. Locally all these builds completed and pushed fine (also with BuildKit). The problem seemingly occurs when writing away the image. I have gone into a lot more detail in the support ticket. I would prefer not to speculate in this PR because the support ticket links directly here.

For now let's await the response from GitHub. 

In the meantime, please feel free to open other pull requests to perform any experiments you might want to try.

PR #20, opened 2020-10-04
Fix hardware ID and tighten the build

webbertakken:
#### Changes

- Remove legacy dependencies
- Upgrade libgtk
- Remove 100 megs from base image
- Fix hardware id for forward compatibility 

#### Checklist

- [X] Readme (updated or not needed)

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -1,5 +1,6 @@
 FROM ubuntu:20.04
 
+# Global dependencies
 RUN apt-get -q update \
  && apt-get -q install -y --no-install-recommends apt-utils \
  && apt-get -q install -y --no-install-recommends --allow-downgrades \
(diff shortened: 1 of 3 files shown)


webbertakken:
Merging this immediately as it also involves some deps, but @jonathanperret could you please review this one retrospectively?

base/Dockerfile
```diff
@@ -1,25 +1,25 @@
...
- zenity \
- libdbus-glib-1-2 \
  curl \
  xz-utils \
  cpio \
  lsb-release \
  && apt-get clean
 
+# Support forward compatibility for unity activation
+RUN echo "576562626572264761624c65526f7578" > /etc/machine-id && ln -sf /etc/machine-id /var/lib/dbus/machine-id
```
GabLeRoux, line 22: Should we use some kind of Docker ARG with a default value here to allow easier customization?

webbertakken: this number should never change, that's the point of it.

Not sure if I understand your meaning.

PR #24, opened 2020-10-15
Added  libgconf-2-4

stamenic:
#### Changes

- Added libgconf-2-4 that was needed

#### Notes

Fixing `/opt/unity/Editor/Unity: error while loading shared libraries: libgconf-2.so.4: cannot open shared object file: No such file or directory`  error.


webbertakken (approved):

webbertakken:
Note that the checks i broke in a commit directly to main while testing dispatcher workflows.

PR #31, opened 2020-10-20
Special tactics

webbertakken:
#### Changes

- Adds exceptions for older versions

#### Checklist

### DIFF
--- a/.github/workflows/new-editor-image-requested.yml
+++ b/.github/workflows/new-editor-image-requested.yml
@@ -17,6 +17,7 @@ jobs:
       fail-fast: false
       matrix:
         targetPlatform:
+          - base # Base is a special name, which indicates that no modules are installed, like with linux-mono on ubuntu
           - linux-il2cpp
           - windows-mono
           - mac-mono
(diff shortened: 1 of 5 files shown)


.github/workflows/new-editor-image-requested.yml
```diff
@@ -38,6 +39,18 @@ jobs:
           echo "repoVersion (full): ${{ github.event.client_payload.repoVersionFull }}"
           echo "repoVersion (only minor and major): ${{ github.event.client_payload.repoVersionMinor }}"
           echo "repoVersion (only major): ${{ github.event.client_payload.repoVersionMajor }}"
+      - name: Special tactics
+        run: |
+          editorVersion="${{ github.event.client_payload.editorVersion }}"
+
+          # For any editor version before 2019.3.0
+          if [[ "$editorVersion" == 201* ]] && [[ "$editorVersion" != 2019.3.* ]] && [[ "$editorVersion" != 2019.4.* ]] ; then
```
GabLeRoux, line 47: Opinion here:  
I think there should be a more _version_ way of comparing them:
See here:
https://gitlab.com/gableroux/unity3d/-/blob/master/ci-generator/src/check_new_version.py#L87

I'm using python's `from packaging import version` and `version.parse` so I can use operators on the versions instead of regex. I feel it's more robust and/or easier to understand. There's probably a way to do something like that with nodejs instead too so its easier to read than bash maybe.

webbertakken: Good idea.

Btw we'll have to go with a different approach, as per the explanation [here](https://github.com/Unity-CI/versioning-backend/pull/23) and the updated pull request (this one).

PR #37, opened 2020-10-26
Added libglu

mashuk999:
#### Changes

- This will solve "error-while-loading-shared-libraries-libglu-so-1" message while running unity-editor.

#### Checklist

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -18,6 +18,7 @@ RUN apt-get -q update \
  wget \
  xvfb \
  xz-utils \
+ libglu1 \
  && apt-get clean
 
 # Support forward compatibility for unity activation


webbertakken:
I'm not too sure what this is needed for. We have confirmation that people are able to run unity-editor without it.

Could you please elaborate which command you're using and what problem it solved for you exactly (except solve the error message)?

BLaZeKiLL:
Got the same error while requesting for the manual activation file

mashuk999:
@BLaZeKiLL `apt-get install libglu1` will solve the issue

mashuk999:
@webbertakken I think in a plain distro the library is not present by default. libglu1 is only 500kb probably adding that will not make much difference i think

webbertakken:
> @webbertakken I think in a plain distro the library is not present by default. libglu1 is only 500kb probably adding that will not make much difference i think

It doesn't. Except that we're republishing half a terabyte in images after this change :) Will merge it after work.

PR #47, opened 2020-11-16
Install Git LFS

bkslasher:
#### Changes

Added installing of Git LFS systemwide. Ubuntu 18's git-lfs is (version 2.3.4-1) is not installed by default after installing the apt package and manual installation is required. Note that an error message is emitted during installation, but it's harmless, see this issue: https://github.com/git-lfs/git-lfs/issues/3013

#### Checklist

- [x] Readme (not needed)

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -30,6 +30,9 @@ RUN apt-get -q update \
  wget \
  && apt-get clean
 
+# Install Git LFS
+RUN git lfs install --system --skip-repo
+
 # Disable default sound card, which removes ALSA warnings
 ADD config/asound.conf /etc/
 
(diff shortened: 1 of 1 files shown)


webbertakken:
According to [this recipe](https://github.com/git-lfs/git-lfs/wiki/Installation#docker-recipes) `git lfs install` should be run before purging apt-get.

Otherwise looks fine to me.

bkslasher:
> According to [this recipe](https://github.com/git-lfs/git-lfs/wiki/Installation#docker-recipes) `git lfs install` should be run before purging apt-get.

Do you mean running it before `apt-get clean` in our case?

webbertakken:
Don't worry about the failing checks. The required ones are green.

Do you think we should adapt the PR or shall we merge it like this?

bkslasher:
Updated

PR #91, opened 2021-03-03
fix: remove cache for building editor image

mob-sakai:
#### Changes

- Remove the cache step in editor image workflows.
- The cache size for GitHub Actions is 5GB, but this is insufficient for caching editor images.
https://github.com/game-ci/docker/runs/2010373210?check_suite_focus=true#step:35:3
> A repository can have up to 5GB of caches. Once the 5GB limit is reached, older caches will be evicted based on when the cache was last accessed. Caches that are not accessed within the last week will also be evicted.
- The cache API increases the build time.
`Cache Docker layers` step takes 2m45s, `Post Cache Docker layers` step takes 2m35s.
https://github.com/game-ci/docker/runs/2010373210?check_suite_focus=true

#### Checklist

### DIFF
--- a/.github/workflows/new-legacy-editor-image-requested.yml
+++ b/.github/workflows/new-legacy-editor-image-requested.yml
@@ -74,16 +74,6 @@ jobs:
             echo "Image already exists. Exiting."
             exit 1
           fi
-      - name: Cache Docker layers
-        uses: actions/cache@v2.1.4
-        with:
-          path: /tmp/.buildx-cache
-          key: ${{ github.event.client_payload.repoVersionFull }}-${{ runner.os }}-buildx-editor-${{ github.event.client_payload.editorVersion }}-${{ matrix.targetPlatform }}-${{ github.sha }}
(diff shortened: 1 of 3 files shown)


webbertakken:
This is a good first iteration, but ideally we'd have the pulled images for base and hub in cache (which were intentionally included in this cache, as per [base](https://github.com/game-ci/docker/blob/main/.github/workflows/new-base-image-requested.yml#L69) and [hub](https://github.com/game-ci/docker/blob/main/.github/workflows/new-hub-image-requested.yml#L69) workflows.

You are entirely correct in that it doesn't work properly now, but hopefully we can at least cache those somehow. Any ideas?

webbertakken (approved):

webbertakken:
Oh we have to make the same changes in [retry-workflow](https://github.com/game-ci/docker/blob/main/.github/workflows/retry-editor-image-requested.yml) as well.

mob-sakai:
I've fix it!

PR #97, opened 2021-03-21
Fix Android environment issues

jsteinich:
#### Changes

- Also update .bashrc in Unity 2018.x
- Actually run the environment updates in Unity 2019+
- Fix issues uncovered when running in Unity 2019+

#### Checklist

- [-] Readme (updated or not needed)

### DIFF
--- a/editor/Dockerfile
+++ b/editor/Dockerfile
@@ -147,7 +147,9 @@ RUN [ `echo $version-$module | grep -vP '^20(?!18).*-android'` ] && exit 0 || : 
   && export RAW_ANDROID_NDK_ROOT=$(jq -cr '(.[] | select(.id == "android-ndk")).destination' $UNITY_PATH/modules.json) \
   && export ANDROID_NDK_HOME=$(echo $RAW_ANDROID_NDK_ROOT | sed -e "s/{UNITY_PATH}/$ESCAPED_UNITY_PATH/g") \
   && export RAW_JAVA_HOME=$(jq -cr '(.[] | select(.id == "android-open-jdk")).destination' $UNITY_PATH/modules.json) \
-  && export JAVA_HOME=$(echo $RAW_JAVA_HOME | sed -e "s/{UNITY_PATH}/$ESCAPED_UNITY_PATH/g") \
+  && export ESCAPED_JAVA_HOME=$(echo $RAW_JAVA_HOME | sed -e "s/{UNITY_PATH}/$ESCAPED_UNITY_PATH/g") \
+  # Unity 2019.x doesn't have the jdk in the modules, so put in a fallback. sdkmanager will fail if invaild
+  && export JAVA_HOME=${ESCAPED_JAVA_HOME:-$UNITY_PATH/Editor/Data/PlaybackEngines/AndroidPlayer/Tools/OpenJDK/Linux} \
   && export PATH=$JAVA_HOME/bin:${ANDROID_SDK_ROOT}/tools:${ANDROID_SDK_ROOT}/tools/bin:${ANDROID_SDK_ROOT}/platform-tools:${PATH} \
   \
   # Clean up
--- a/editor/Dockerfile
+++ b/editor/Dockerfile
@@ -120,6 +120,7 @@ RUN [ `echo $version-$module | grep -v '^\(2018.3\|2018.4\).*-android'` ] && exi
     echo 'xvfb-run -ae /dev/stdout "$UNITY_PATH/Editor/Unity" -batchmode "$@"'; \
   } > /usr/bin/unity-editor \
   && chmod +x /usr/bin/unity-editor \
+  # Update '~/.bashrc' to enable using variables when logging in
   && { \
     echo "export ANDROID_SDK_ROOT=${ANDROID_SDK_ROOT}"; \
     echo "export ANDROID_HOME=${ANDROID_HOME}"; \
@@ -139,6 +140,7 @@ RUN [ `echo $version-$module | grep -vP '^20(?!18).*-android'` ] && exit 0 || : 
   \
   # Environment Variables
   && export RAW_ANDROID_SDK_ROOT=$(jq -cr '(.[] | select(.id == "android-sdk-platform-tools")).destination' $UNITY_PATH/modules.json) \
+  # We need to replace some characters common to paths that will break the sed expression when expanded
   && export ESCAPED_UNITY_PATH=$(printf '%s' "$UNITY_PATH" | sed 's/[#\/]/\\\0/g') \
   && export ANDROID_SDK_ROOT=$(echo $RAW_ANDROID_SDK_ROOT | sed -e "s/{UNITY_PATH}/$ESCAPED_UNITY_PATH/g") \
   && export ANDROID_HOME=${ANDROID_SDK_ROOT} \
@@ -149,6 +151,7 @@ RUN [ `echo $version-$module | grep -vP '^20(?!18).*-android'` ] && exit 0 || : 
   && export PATH=$JAVA_HOME/bin:${ANDROID_SDK_ROOT}/tools:${ANDROID_SDK_ROOT}/tools/bin:${ANDROID_SDK_ROOT}/platform-tools:${PATH} \
   \
   # Clean up
+  && apt-get remove -y jq \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/* \
   \
@@ -168,6 +171,7 @@ RUN [ `echo $version-$module | grep -vP '^20(?!18).*-android'` ] && exit 0 || : 
     echo 'xvfb-run -ae /dev/stdout "$UNITY_PATH/Editor/Unity" -batchmode "$@"'; \
   } > /usr/bin/unity-editor \
   && chmod +x /usr/bin/unity-editor \
+  # Update '~/.bashrc' to enable using variables when logging in
   && { \
     echo "export ANDROID_SDK_ROOT=${ANDROID_SDK_ROOT}"; \
     echo "export ANDROID_HOME=${ANDROID_HOME}"; \
--- a/editor/Dockerfile
+++ b/editor/Dockerfile
@@ -119,21 +119,38 @@ RUN [ `echo $version-$module | grep -v '^\(2018.3\|2018.4\).*-android'` ] && exi
     echo ''; \
     echo 'xvfb-run -ae /dev/stdout "$UNITY_PATH/Editor/Unity" -batchmode "$@"'; \
   } > /usr/bin/unity-editor \
-  && chmod +x /usr/bin/unity-editor
+  && chmod +x /usr/bin/unity-editor \
+  && { \
+    echo "export ANDROID_SDK_ROOT=${ANDROID_SDK_ROOT}"; \
+    echo "export ANDROID_HOME=${ANDROID_HOME}"; \
+    echo "export ANDROID_NDK_HOME=${ANDROID_NDK_HOME}"; \
+    echo "export JAVA_HOME=${JAVA_HOME}"; \
+    echo "export PATH=${PATH}"; \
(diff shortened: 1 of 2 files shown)


editor/Dockerfile
```diff
@@ -119,22 +119,39 @@ RUN [ `echo $version-$module | grep -v '^\(2018.3\|2018.4\).*-android'` ] && exi
...
+    echo "export ANDROID_NDK_HOME=${ANDROID_NDK_HOME}"; \
+    echo "export JAVA_HOME=${JAVA_HOME}"; \
+    echo "export PATH=${PATH}"; \
+  } > ~/.bashrc
 
 #=======================================================================================
 # [2019.x/2020.x/2021.x-android] Support usage of installed Android SDKs outside of editor
 #=======================================================================================
-RUN [ `echo $version-$module | grep -v '^20(?!18).*-android'` ] && exit 0 || : \
+RUN [ `echo $version-$module | grep -vP '^20(?!18).*-android'` ] && exit 0 || : \
```
jsteinich, line 134: Need `-P` to enable negative lookahead support

editor/Dockerfile
```diff
@@ -119,22 +119,39 @@ RUN [ `echo $version-$module | grep -v '^\(2018.3\|2018.4\).*-android'` ] && exi
...
+  \
+  # Install jq
+  && apt-get update -qq \
+  && apt-get install -qq -y --no-install-recommends jq \
```
jsteinich, line 138: Actually install `jq` which is used to parse Unity `modules.json` file.

editor/Dockerfile
```diff
@@ -119,22 +119,39 @@ RUN [ `echo $version-$module | grep -v '^\(2018.3\|2018.4\).*-android'` ] && exi
...
   \
   # Environment Variables
   && export RAW_ANDROID_SDK_ROOT=$(jq -cr '(.[] | select(.id == "android-sdk-platform-tools")).destination' $UNITY_PATH/modules.json) \
-  && export ANDROID_SDK_ROOT=${RAW_ANDROID_SDK_ROOT/'{UNITY_PATH}'/$UNITY_PATH} \
+  && export ESCAPED_UNITY_PATH=$(printf '%s' "$UNITY_PATH" | sed 's/[#\/]/\\\0/g') \
+  && export ANDROID_SDK_ROOT=$(echo $RAW_ANDROID_SDK_ROOT | sed -e "s/{UNITY_PATH}/$ESCAPED_UNITY_PATH/g") \
```
jsteinich, line 143: Since we aren't in a full bash environment, jump through some sed hoops to do the replacement.

editor/Dockerfile
```diff
@@ -119,22 +119,39 @@ RUN [ `echo $version-$module | grep -v '^\(2018.3\|2018.4\).*-android'` ] && exi
...
+  && export ANDROID_HOME=${ANDROID_SDK_ROOT} \
```
jsteinich, line 144: More consistent with 2018.x block

editor/Dockerfile
```diff
@@ -119,22 +119,39 @@ RUN [ `echo $version-$module | grep -v '^\(2018.3\|2018.4\).*-android'` ] && exi
...
   && export RAW_ANDROID_NDK_ROOT=$(jq -cr '(.[] | select(.id == "android-ndk")).destination' $UNITY_PATH/modules.json) \
-  && export ANDROID_NDK_ROOT=${RAW_ANDROID_NDK_ROOT/'{UNITY_PATH}'/$UNITY_PATH} \
+  && export ANDROID_NDK_HOME=$(echo $RAW_ANDROID_NDK_ROOT | sed -e "s/{UNITY_PATH}/$ESCAPED_UNITY_PATH/g") \
```
jsteinich, line 146: More consistent with 2018.x block

jsteinich:
Looks like 2019.1 doesn't have openjdk in the modules.json file. I found it at `/opt/unity/Editor/Data/PlaybackEngines/AndroidPlayer/Tools/OpenJDK/Linux`. Considering just hardcoding that path for 2019.1, but I'm open to other suggestions.

editor/Dockerfile
```diff
@@ -119,22 +119,39 @@ RUN [ `echo $version-$module | grep -v '^\(2018.3\|2018.4\).*-android'` ] && exi
     echo ''; \
     echo 'xvfb-run -ae /dev/stdout "$UNITY_PATH/Editor/Unity" -batchmode "$@"'; \
   } > /usr/bin/unity-editor \
-  && chmod +x /usr/bin/unity-editor
+  && chmod +x /usr/bin/unity-editor \
+  && { \
```
webbertakken, line 123: Could you also add a comment before this block?

editor/Dockerfile
webbertakken: I have never seen `jq` used by Unity, do you have a source for this?

If this is used to parse Unity `modules.json`, should it not be installed for all versions, instead of making it part of the android specific fixes for 2019.x through 2021.x?

And if so, we should probably add it to the base image instead of the editor image, so we can also use it in Hub.

In case we're just adding it as a nice to have utility without Unity Editor actually needing it, we should check what the size of the installation with `--no-recommends` is. (e.g. 10MB+ just for parsing json would not be useful for CI images).

editor/Dockerfile
webbertakken: You seem to know a lot about bash scripting. 🚀

Do you have recommended reading for me to learn more about what it means to have a "full bash environment" and how to find out? 

I think I'm having a knowledge gap there.

editor/Dockerfile
webbertakken: In 2018 block it looks like this

```
&& export ANDROID_NDK_HOME=${ANDROID_SDK_ROOT}/ndk/${ANDROID_NDK_VERSION} \
```

Did you forget to add the ndk/{version} part or is this correct? (this is hard to check for me)

editor/Dockerfile
```diff
@@ -119,22 +119,39 @@ RUN [ `echo $version-$module | grep -v '^\(2018.3\|2018.4\).*-android'` ] && exi
...
+RUN [ `echo $version-$module | grep -vP '^20(?!18).*-android'` ] && exit 0 || : \
+  \
+  # Install jq
+  && apt-get update -qq \
+  && apt-get install -qq -y --no-install-recommends jq \
   \
   # Environment Variables
   && export RAW_ANDROID_SDK_ROOT=$(jq -cr '(.[] | select(.id == "android-sdk-platform-tools")).destination' $UNITY_PATH/modules.json) \
-  && export ANDROID_SDK_ROOT=${RAW_ANDROID_SDK_ROOT/'{UNITY_PATH}'/$UNITY_PATH} \
+  && export ESCAPED_UNITY_PATH=$(printf '%s' "$UNITY_PATH" | sed 's/[#\/]/\\\0/g') \
```
webbertakken, line 142: Could you add a comment above this line to explain what this replacement is doing exactly?

editor/Dockerfile
webbertakken: I think this is why the variable was called NDK_ROOT and not NDK_HOME, although I'm not sure.

@mob-sakai could you please have a look as well?

webbertakken:
> Looks like 2019.1 doesn't have openjdk in the modules.json file. I found it at `/opt/unity/Editor/Data/PlaybackEngines/AndroidPlayer/Tools/OpenJDK/Linux`. Considering just hardcoding that path for 2019.1, but I'm open to other suggestions.

Do we know if this is limited to 2019.1.x?

As far as we know these tools are installed based on the json file that the hub retrieves based on the version number, which is I believe the same file that gets placed in the editor folder.

We can add it hardcoded, but we'll need to show and verify that this actually works.

Reason for this is that in the previous docker images from `gableroux/unity3d` we added some things that turned out to be not 100% correct, and we're trying to learn from this mistake, and be quite strict about testing and verifying that changes indeed make sense.

webbertakken:
Thank you @jsteinich, these look like some really great improvements! 🚀

Some of the changes need to be provided with rationale, so that we merge changes as objectively as we can.

Looking forward to merging this!

editor/Dockerfile
jsteinich: It is currently only used by that particular android setup block. I'll just remove it after use.

editor/Dockerfile
jsteinich: I'm no expert. I just know enough to let Google fill in the blanks.

Docker run uses a login less shell which defaults to `/bin/sh` which actually symlinks to `dash`. Logging in normally will default to `bash`.
Here is a simple breakdown of the [2 shells](https://blog.cloudware.bg/en/dash-vs-bash-shell/).

editor/Dockerfile
jsteinich: When installing the NDK with the sdkmanager, it places it into a versioned folder so that you can easily have multiple versions installed. Unity simply installs a single version which it places into a generic `NDK` folder.

jsteinich:
> Do we know if this is limited to 2019.1.x?

I'm not certain. I only found out about it because the test for that version failed.

I haven't really dug into how the tests for this project work, but perhaps we could add some reporting to what the environment variable values are and/or if the tools are useable.

jsteinich:
All 2018.3.14f1 tests are failing with `Assembly has reference to non-existent assembly 'Unity.Addressables.Editor' (Assets/Editor/Editor/UnityBuilderAction/UnityBuilderAction.asmdef)`. I don't believe that is related to this PR. Instead, I'm thinking it is caused by https://github.com/game-ci/unity-builder/pull/231

webbertakken:
> All 2018.3.14f1 tests are failing with `Assembly has reference to non-existent assembly 'Unity.Addressables.Editor' (Assets/Editor/Editor/UnityBuilderAction/UnityBuilderAction.asmdef)`. I don't believe that is related to this PR. Instead, I'm thinking it is caused by [game-ci/unity-builder#231](https://github.com/game-ci/unity-builder/pull/231)

Yes. This is not related to your PR indeed.

@sbergen any ideas on how to fix this?

sbergen:
> > All 2018.3.14f1 tests are failing with `Assembly has reference to non-existent assembly 'Unity.Addressables.Editor' (Assets/Editor/Editor/UnityBuilderAction/UnityBuilderAction.asmdef)`. I don't believe that is related to this PR. Instead, I'm thinking it is caused by [game-ci/unity-builder#231](https://github.com/game-ci/unity-builder/pull/231)
> 
> Yes. This is not related to your PR indeed.
> 
> @sbergen any ideas on how to fix this?

Hmm, I'm thinking that they only made that not-an-error when they introduced version defines in 2019. Didn't think of this 😞 

Unfortunately, you can't have anything like `#if UNITY_2019...` in an .asmdef, so I can't think of a really elegant solution.

The options I can think of are:
* Making two versions of the .asmdef: one for >= 2019, and another for older versions
* Reverting back to not using an asmdef, and making that logic be behind a builder option, instead of using version defines. You'd then define some symbol from the build script for whether or not the logic should be included.
* Parsing `manifest.json` manually, and using that to define a symbol, instead of doing it based on an option.

webbertakken:
@sbergen I have created a bug report for it. https://github.com/game-ci/unity-builder/issues/233

We can discuss a solution there and someone might be able to implement it.

jsteinich:
I added in a step to the test workflow that prints out the java version, sdkmanager version, and ndk version. It's by no means an exhaustive test, but should give some indication on whether or not java, android sdk, and android ndk are installed correctly without maintaining a map of expected versions.

jsteinich:
@webbertakken the only failing test at this point is 2021.1.3f1, linux-il2cpp which seems completely unrelated. Is this good to merge, or are there more changes you'd like made?

webbertakken:
Since the exhausted memory error only occurs in 2021.1 I believe it is indeed not related.

webbertakken:
Great work, thank you very much for your contribution

PR #106, opened 2021-04-28
Adds `runner` user in hub/Dockerfile

Bradshaw:
#### Changes

- Adds a `runner` user with uid:gid `1000:1000` in `hub/Dockerfile`
- This is intended to allow downstream steps to run as this user and not create root-owned files

*c.f:* [game-ci/unity-builder issue #219](https://github.com/game-ci/unity-builder/issues/219)

#### Explanation

(From comment: https://github.com/game-ci/docker/pull/106#issuecomment-830375895)

Here's what we're doing:

In `base/Dockerfile`:
- [Add `UNAME`, `UID` and `GID`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L3-L9) that we use to parameterise our non-root user. By default: `runner 1000:1000`.
- [Install `sudo`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L35)
- [Create the runner user](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L57-L59)
- [`chown` key executable folders to `runner`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L61-L63) (this lets us write to executable folders)
- [Let `runner` use `sudo apt-get` and `sudo rm` without a password](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L65-L66)
- [And then switch to user `runner`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L68-L69)

In `hub/Dockerfile` (now running as `runner 1000:1000`)
- [Use `sudo apt-get`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/hub/Dockerfile#L5-L7) (we won't be bothered for a password)
- [Install Unity files to `$UNITY_PATH`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/hub/Dockerfile#L9-L17) (instead of `/`)
- [Use `APPIMAGE_SILENT_INSTALL` and `APPDIR` env vars in `/usr/bin/unity-hub`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/hub/Dockerfile#L19-L21) (Silent install skips EULA check, appdir lets the UnityHub script know where to look for Unity files)

In `editor/Dockerfile`:
- Use `sudo apt-get` and `sudo rm` a whole bunch 🤷

All of these are now running as `runner` (or whatever you pass as `UNAME` build parameter.

We don't *need* to have `UNAME`, `UID` and `GID` copied to env vars, but it might be useful if dependants want to `USER 0:0` to run some stuff and be able to `USER $UID:$GID` back...

#### Caveats and other information

- This breaks pipelines that depend on root user (including [game-ci/unity-builder](https://github.com/game-ci/unity-builder). If they want to use these images, they need to add a `USER root` command, or not rely on root any more
- If the runner user id and group id don't exist on the host system, it can cause some issues too, since the files will belong to an invalid user and group.
- It is possible to run images as a custom user (*e.g:* `docker run -u 1337:4242`), but there can be issues if the user doesn't exist in the image. It's possible to solve this by creating an image that imports these images and creating a user there before continuing.

#### Checklist

- [x] Readme not needed
- [ ] Update tests to check that the generated files belong to `runner` after building the test project

### DIFF
--- a/base/Dockerfile
+++ b/base/Dockerfile
@@ -1,5 +1,13 @@
 FROM ubuntu:18.04
 
+# Runner user parameters
+ARG UNAME=runner
+ENV UNAME=$UNAME
+ARG UID=1000
+ENV UID=$UID
+ARG GID=1000
+ENV GID=$GID
+
 # Global dependencies
 RUN apt-get -q update \
  && apt-get -q install -y --no-install-recommends apt-utils \
@@ -24,6 +32,7 @@ RUN apt-get -q update \
 # Toolbox
 RUN apt-get -q update \
  && apt-get -q install -y --no-install-recommends --allow-downgrades \
+ sudo \
  atop \
  curl \
  git \
@@ -44,3 +53,17 @@ ENV UNITY_PATH="/opt/unity"
 # Fixes a Gradle crash while building for Android on Unity 2019 when there are accented characters in environment variables
 ENV LANG C.UTF-8
 ENV LC_ALL C.UTF-8
+
+# Create the runner user
+RUN groupadd -g $GID $UNAME
+RUN useradd -u $UID -g $UNAME -s /bin/sh -m $UNAME
+
+# Give relevant executable folders to runner user
+RUN mkdir -p $UNITY_PATH && chown $UID:$GID $UNITY_PATH
+RUN chown $UID:$GID /usr/bin
+
+# Let the runner user use apt-get and rm without a password
+RUN echo ${UNAME}" ALL = NOPASSWD : /bin/rm , /usr/bin/apt-get" > /etc/sudoers.d/apt
+
+# Run as runner user for the rest of the steps
+USER $UID:$GID
--- a/hub/Dockerfile
+++ b/hub/Dockerfile
@@ -2,29 +2,25 @@ ARG baseImage="unityci/base"
 FROM $baseImage
 
 # Hub dependencies
-RUN apt-get -q update \
- && apt-get -q install -y --no-install-recommends --allow-downgrades zenity \
- && apt-get clean
+RUN sudo apt-get -q update \
+ && sudo apt-get -q install -y --no-install-recommends --allow-downgrades zenity \
+ && sudo apt-get clean
 
 # Download & extract AppImage
 RUN wget --no-verbose -O /tmp/UnityHub.AppImage "https://public-cdn.cloud.unity3d.com/hub/prod/UnityHub.AppImage" \
  && chmod +x /tmp/UnityHub.AppImage \
  && cd /tmp \
  && /tmp/UnityHub.AppImage --appimage-extract \
- && cp -R /tmp/squashfs-root/* / \
- && rm -rf /tmp/squashfs-root /tmp/UnityHub.AppImage \
  && mkdir -p "$UNITY_PATH" \
- && mv /AppRun /opt/unity/UnityHub
+ && cp -R /tmp/squashfs-root/* $UNITY_PATH \
+ && rm -rf /tmp/squashfs-root /tmp/UnityHub.AppImage \
+ && mv $UNITY_PATH/AppRun $UNITY_PATH/UnityHub
 
 # Alias to "unity-hub" with default params
-RUN echo '#!/bin/bash\nxvfb-run -ae /dev/stdout /opt/unity/UnityHub --no-sandbox --headless "$@"' > /usr/bin/unity-hub \
+RUN echo '#!/bin/bash\nAPPIMAGE_SILENT_INSTALL=true APPDIR='$UNITY_PATH' xvfb-run -ae /dev/stdout '$UNITY_PATH'/UnityHub --no-sandbox --headless "$@"' > /usr/bin/unity-hub \
  && chmod +x /usr/bin/unity-hub
-
-# Accept
-RUN mkdir -p "/root/.config/Unity Hub" \
- && touch "/root/.config/Unity Hub/eulaAccepted"
 
 # Configure
 RUN mkdir -p "${UNITY_PATH}/editors" \
  && unity-hub install-path --set "${UNITY_PATH}/editors/" \
- && find /tmp -mindepth 1 -delete
+ && find /tmp -mindepth 1 -delete 2> /dev/null
--- a/hub/Dockerfile
+++ b/hub/Dockerfile
@@ -17,7 +17,7 @@ RUN wget --no-verbose -O /tmp/UnityHub.AppImage "https://public-cdn.cloud.unity3
  && mv $UNITY_PATH/AppRun $UNITY_PATH/UnityHub
 
 # Alias to "unity-hub" with default params
-RUN echo '#!/bin/bash\nAPPIMAGE_SILENT_INSTALL=true APPDIR='$UNITY_PATH' xvfb-run -ae /dev/stdout '$UNITY_PATH'/UnityHub --no-sandbox --headless "$@"' > /usr/bin/unity-hub \
+RUN echo '#!/bin/bash\nAPPIMAGE_SILENT_INSTALL=true APPDIR=$UNITY_PATH xvfb-run -ae /dev/stdout $UNITY_PATH/UnityHub --no-sandbox --headless "$@"' > /usr/bin/unity-hub \
  && chmod +x /usr/bin/unity-hub
 
 # Configure
--- a/editor/Dockerfile
+++ b/editor/Dockerfile
@@ -29,7 +29,7 @@ FROM $baseImage
 # Always put "Editor" and "modules.json" directly in $UNITY_PATH
 ARG version
 ARG module
-COPY --from=builder /opt/unity/editors/$version/ "$UNITY_PATH/"
+COPY --from=builder $UNITY_PATH/editors/$version/ "$UNITY_PATH/"
 
 # Add a file containing the version for this build
 RUN echo $version > "$UNITY_PATH/version"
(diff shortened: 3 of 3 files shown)


hub/README.md
```diff
@@ -9,7 +9,7 @@ Dockerised Unity Hub made for continuous integration.
 Run the hub image using an interactive shell
 
 ```bash
-docker run -it --rm unityci/hub bash
+docker run -it --rm gaeelbradshaw/hub bash
```
GabLeRoux, line 12: this will need to change :)

hub/Dockerfile
```diff
@@ -1,6 +1,12 @@
-ARG baseImage="unityci/base"
+ARG baseImage="gaeelbradshaw/base"
```
GabLeRoux, line 1: :)

Bradshaw:
I'm fairly sure the single failed test is just a good old timeout
However, this does not fix the issue, since the editor image is still running as root, @webbertakken has explained the base->hub / base->editor model to me on Discord and I've realised I have some more work to do

Bradshaw:
I'm currently solving the EULA problem by setting a `APPIMAGE_SILENT_INSTALL` environment variable. It means there's no need to create a `eulaAccepted` file in the correct folder.

```Dockerfile
RUN echo '#!/bin/bash\nAPPIMAGE_SILENT_INSTALL=true APPDIR='$APPDIR' xvfb-run -ae /dev/stdout '$UNITY_PATH'/UnityHub --no-sandbox --headless "$@"' > $RUNNER_HOME/bin/unity-hub \
 && chmod +x $RUNNER_HOME/bin/unity-hub
```

I feel like this is equivalent to the previous approach, since both are transparent to the user and effectively run the software without a human accepting the EULA, I don't know enough about the specifics of licencing law to be 100% certain though.

hub/Dockerfile
```diff
@@ -2,29 +2,25 @@ ARG baseImage="unityci/base"
...
 
-# Accept
-RUN mkdir -p "/root/.config/Unity Hub" \
- && touch "/root/.config/Unity Hub/eulaAccepted"
-
 # Configure
 RUN mkdir -p "${UNITY_PATH}/editors" \
  && unity-hub install-path --set "${UNITY_PATH}/editors/" \
- && find /tmp -mindepth 1 -delete
+ && find /tmp -mindepth 1 -delete 2> /dev/null
```
webbertakken, line 26: Why is this needed?

hub/Dockerfile
```diff
@@ -2,29 +2,25 @@ ARG baseImage="unityci/base"
...
- && rm -rf /tmp/squashfs-root /tmp/UnityHub.AppImage \
  && mkdir -p "$UNITY_PATH" \
- && mv /AppRun /opt/unity/UnityHub
+ && cp -R /tmp/squashfs-root/* $UNITY_PATH \
+ && rm -rf /tmp/squashfs-root /tmp/UnityHub.AppImage \
+ && mv $UNITY_PATH/AppRun $UNITY_PATH/UnityHub
 
 # Alias to "unity-hub" with default params
-RUN echo '#!/bin/bash\nxvfb-run -ae /dev/stdout /opt/unity/UnityHub --no-sandbox --headless "$@"' > /usr/bin/unity-hub \
+RUN echo '#!/bin/bash\nAPPIMAGE_SILENT_INSTALL=true APPDIR='$UNITY_PATH' xvfb-run -ae /dev/stdout '$UNITY_PATH'/UnityHub --no-sandbox --headless "$@"' > /usr/bin/unity-hub \
```
webbertakken, line 20: Quote style that we're using everywhere is 

```bash
"PARAM=$SOME_VAR"
```
or 
```bash
"${SOME_VAR}suffix"
```
or
```bash
"${UNITY_PATH}/editors/"
```

single quotes is fine too. 

just not 

```bash
'$UNITY_PATH'/UnityHub
```

Especially because it's using quotes around _part of the string_, within a string that's _already quoted in single quotes_ as well.

webbertakken:
LGTM

If this works we can move forward I think.

Bradshaw:
I think I got it!

Here's what we're doing:

In `base/Dockerfile`:
- [Add `UNAME`, `UID` and `GID`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L3-L9) that we use to parameterise our non-root user. By default: `runner 1000:1000`.
- [Install `sudo`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L35)
- [Create the runner user](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L57-L59)
- [`chown` key executable folders to `runner`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L61-L63) (this lets us write to executable folders)
- [Let `runner` use `sudo apt-get` and `sudo rm` without a password](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L65-L66)
- [And then switch to user `runner`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/base/Dockerfile#L68-L69)

In `hub/Dockerfile` (now running as `runner 1000:1000`)
- [Use `sudo apt-get`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/hub/Dockerfile#L5-L7) (we won't be bothered for a password)
- [Install Unity files to `$UNITY_PATH`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/hub/Dockerfile#L9-L17) (instead of `/`)
- [Use `APPIMAGE_SILENT_INSTALL` and `APPDIR` env vars in `/usr/bin/unity-hub`](https://github.com/Bradshaw/docker/blob/f5f67e561ebb2500b6a6663ef8cb9645795d482f/hub/Dockerfile#L19-L21) (Silent install skips EULA check, appdir lets the UnityHub script know where to look for Unity files)

In `editor/Dockerfile`:
- Use `sudo apt-get` and `sudo rm` a whole bunch 🤷

All of these are now running as `runner` (or whatever you pass as `UNAME` build parameter.

We don't *need* to have `UNAME`, `UID` and `GID` copied to env vars, but it might be useful if dependants want to `USER 0:0` to run some stuff and be able to `USER $UID:$GID` back...

hub/Dockerfile
Bradshaw: Simple because as a non-root, deleting `/tmp` files prints an error because we're also trying to delete files created by root...

We can probably do without tbh

hub/Dockerfile
Bradshaw: The problem here is that we're echoing a long string to a text file, and we need ARGs to be passed in...
Maybe I'm confused as to the string interpolation syntax interactions between docker and bash, but I'll take a look

I wouldn't mind a recommendation TBH

webbertakken: you can just quote args like the last arg in that sentence: `"$@"`

so `'$UNITY_PATH'/UnityHub` becomes `"$UNITY_PATH/UnityHub"` 

and `APPDIR='$UNITY_PATH'` becomes `"APPDIR=$UNITY_PATH"`

base/Dockerfile
```diff
@@ -1,5 +1,13 @@
 FROM ubuntu:18.04
 
+# Runner user parameters
+ARG UNAME=runner
+ENV UNAME=$UNAME
```
webbertakken, line 5: About UNAME, I think it should be USER or USERNAME right? Currently this conflicts with [uname](https://man7.org/linux/man-pages/man2/uname.2.html)

base/Dockerfile
```diff
@@ -1,5 +1,13 @@
...
+ARG UID=1000
+ENV UID=$UID
+ARG GID=1000
+ENV GID=$GID
```
webbertakken, line 9: Do we have to expose these args or can we just use ENV only for these?

hub/Dockerfile
Bradshaw: This doesn't work, because we're in a single-quote string

In the case of the `"$@" ` we actually want `"$@"` in the body of our `unity-hub` script
Those single quotes are actually closing and opening, not opening and closing

But it turns out they're not needed at all

base/Dockerfile
Bradshaw: If we only keep one or the other, it's better to keep ARG, this is a build-time use-case, and only `base` actually needs it

I kept the ENV in case people want to be able to do something like
```Dockerfile
USER 0:0
RUN some --root-commands
USER $UID:$GID
RUN cat regular-user.file
```

hub/Dockerfile
webbertakken: Let's not touch existing quotes. They originate from some original script.

I don't see how they're closing and opening instead or vice versa by the way.

I guess the newly added quotes are not necessary

lastthyme (approved):

Bradshaw:
I don't think this should be approved yet, @lastthyme, there's still conversation ongoing in the Discord, and we might fix this in a completely different way

webbertakken:
Note that the file permissions issue / files owned by root, is a github specific issue.

https://github.community/t/actions-files-changed-in-container-action-are-now-owned-by-root/16115/3

Bradshaw:
> Note that the file permissions issue / files owned by root, is a github specific issue.
> 
> https://github.community/t/actions-files-changed-in-container-action-are-now-owned-by-root/16115/3

Those comments are from *2019*! It doesn't look like GitHub is interested in fixing the issue any time soon

webbertakken:
This currently does not look viable. I'm closing this PR until there is more actionable information.

PR #116, opened 2021-05-26
Allow to install multiple modules in the editor image

shr-project:
#### Changes

- Instead of having separate image for windows, android, mac, ... you can now create just one image which will have all modules installed.
- Existing module build-arg now allows to use space separated list of modules.

#### Checklist


editor/Dockerfile
```diff
@@ -132,8 +138,9 @@ RUN [ `echo $version-$module | grep -v '^\(2018.3\|2018.4\).*-android'` ] && exi
 #=======================================================================================
 # [2019.x/2020.x/2021.x-android] Support usage of installed Android SDKs outside of editor
 #=======================================================================================
-RUN [ `echo $version-$module | grep -vP '^20(?!18).*-android'` ] && exit 0 || : \
-  \
+RUN echo "$version-$module" | grep -q -vP '^20(?!18).*android' \
+  && exit 0 \
+  || : \
```
webbertakken, line 143: Note that this will currently not work, as one file will overwrite the other when multiple components are installed.

editor/Dockerfile
```diff
@@ -10,15 +10,17 @@ FROM $hubImage AS builder
 # Install editor
 ARG version
 ARG changeSet
-RUN unity-hub install --version "$version" --changeset "$changeSet" | grep 'Error' | exit $(wc -l)
+RUN unity-hub install --version "$version" --changeset "$changeSet" | tee install.log && grep 'Error' install.log | exit $(wc -l)
```
webbertakken, line 13: Great idea to keep the install log in a file.

I would suggest using `/var/log/install-editor.log` and `/var/log/install-editor-module-${mod}.log` instead, so that there are no floating files in the root.

editor/Dockerfile
```diff
@@ -10,15 +10,17 @@ FROM $hubImage AS builder
...
 # Install modules for that editor
 ARG module="non-existent-module"
-RUN if [ "$module" = "base" ] ; then \
-      echo "running default modules for this baseOs" ; exit 0 ; \
-    else \
-      unity-hub install-modules --version "$version" --module "$module" --childModules | grep 'Missing module' | exit $(wc -l) ; \
-    fi
+RUN for mod in $module; do \
+      if [ "$module" = "base" ] ; then \
+        echo "running default modules for this baseOs" ; exit 0 ; \
```
webbertakken, line 19: `exit 0` is no longer appropriate here. I think we can safely remove it.

webbertakken (changes requested):
Thanks for your proposal!

I think this is worth merging, because specific builds require more than one module. I have left a few comments that would need to be addressed before doing so.

Please also note that we generally discourage our users from installing all modules in one single image as it goes against parallelisation in many cases. In many cases pulling the (compressed) image directly is just as fast as caching, however when trying to cache an image that holds all modules, it will actually slow down each build (even when cached, depending on the mechanism).

editor/Dockerfile
shr-project: Can you explain a bit more? I haven't tried all possible combinations, but quickly reading through them I don't see where they would conflict with each other. You will still install just one version of android so only one of these section should run even when you install e.g. windows-mono and mac-mono together with android.

But it's completely possible that I've overlooked something, I was testing this only with 2020.3.3f1 version and "windows-mono android mac-mono" modules.

shr-project:
> Please also note that we generally discourage our users from installing all modules in one single image as it goes against parallelisation in many cases. In many cases pulling the (compressed) image directly is just as fast as caching, however when trying to cache an image that holds all modules, it will actually slow down each build (even when cached, depending on the mechanism).

Understood, I was going to use this for building SVL Simulator (https://github.com/lgsvl/simulator/blob/master/Jenkins/Jenkinsfile) where the initial checkout stages takes quite long, so instead of triggering 3 jobs for 3 different platforms in parallel it builds all 3 platforms in 1 job and the unity version isn't changed very often, so the bigger editor docker image can stay cached on the builders for relatively long.

Currently it was using UnitySetup instead of UnityHub for installation (with linux, windows, mac support all included):
https://github.com/lgsvl/simulator/blob/master/Jenkins/Dockerfile#L26
but I haven't found a way how to let UnitySetup install childModules as well (which seems to be mostly useful for android).

editor/Dockerfile
webbertakken: Almost each conditional section writes to the alias file `unity-editor` and some to `~/.bashrc`.

However with your approach we'll need a smart way to combine some of this logic and prevent overwriting the .bashrc with each module.

shr-project: But each of these sections is for different version and only one of them will be executed for each docker image, right?

webbertakken: There is only 1 file being created, but it's being created it multiple places.

On line 68, 128, 185.

The code so far was not made to scale for multiple modules. For example, if you were to use webgl or linux-il2cpp combined with android, the `webgl` and `linux-il2cpp` specific fixes are being undone.

What we need to make this PR merge-able is to find an elegant way account for multiple modules when creating that alias script (and the .bashrc file).

shr-project: See the proposed changes.

DEVELOPMENT.md
```diff
@@ -17,19 +17,19 @@ cd docker
...
 ```bash
-docker build hub -t hub
+docker build -t hub -f hub/Dockerfile .
 ```
 
 Build editor
 
 ```bash
-docker build editor -t editor
+docker build -t editor -f editor/Dockerfile .
```
webbertakken, line 32: We actually are planning to move away from this context in an upcoming change. But it's ok to change it for now. Just fyi.

editor/Dockerfile
```diff
@@ -39,28 +41,30 @@ RUN echo $version > "$UNITY_PATH/version"
 #  Alias to unity-editor  #
 ###########################
 
-RUN echo '#!/bin/bash' > /usr/bin/unity-editor \
-  && chmod +x /usr/bin/unity-editor
+ADD editor/unity-editor /usr/bin/unity-editor
+RUN chmod 755 /usr/bin/unity-editor \
+  && mkdir /usr/bin/unity-editor.d
```
webbertakken, line 46: Adding this external file and running different .sh files inside a folder is adding a lot of unnecessary complexity.

In my opinion the best way to solve the file overwrites would have been to simply move the last line of that file after the android clauses, and make sure that each addition is only added once.

I'm sorry about having to request changes again. It is however quite important that this Dockerfile stays very clean. As we publish thousands of images with it.

editor/Dockerfile
```diff
@@ -88,7 +93,7 @@ RUN [ `echo $version-$module | grep -v '^\(2018.3\|2018.4\).*-android'` ] && exi
   && apt-get update -qq \
   && apt-get install -qq -y --no-install-recommends unzip \
   && mkdir -p ${ANDROID_SDK_ROOT} \
-  && chown -R 777 ${ANDROID_INSTALL_LOCATION} \
+  && chmod -R 777 ${ANDROID_INSTALL_LOCATION} \
```
webbertakken, line 96: Was this in fact giving an error? If not, it would be better to make separate PRs for changes that are not related to supporting multiple modules.

editor/Dockerfile
shr-project: But it also makes it useful for extending this script if someone needs to export few more variables in unity-editor script in another docker image which will be based on this editor.

Every /etc/profile does this with /etc/profile.d/*sh and many other scripts/configs use the same mechanism to combine various pieces together.

editor/Dockerfile
shr-project: I didn't notice the error, but I've noticed this when adding that chmod for sdkmanager.

editor/Dockerfile
webbertakken: This is out of scope for this change. 

There are much easier ways of adding variables.

When using Unity Builder, you can add it specifically for that workflow in its extended [Dockerfile](https://github.com/game-ci/unity-builder/tree/main/dist).

When not using Builder, it's still very easy to achieve by using

```dockerfile
FROM <this image:some-version-with-specific-compatibility-fixes-maintained-by-game-ci>

ENV my_var="my var"
```

editor/Dockerfile
shr-project: separated into https://github.com/game-ci/docker/pull/119

DEVELOPMENT.md
shr-project: OK, I just noticed that this didn't work (because of config/asound.conf in base/Dockerfile) and wanted to adjust the documentation before adding the same assumption about context with editor/unity-editor file.

editor/Dockerfile
shr-project: > I would argue that this is out of scope for this proposal.

Well, you wanted an elegant way in:

> What we need to make this PR merge-able is to find an elegant way account for multiple modules when creating that unity-editor alias script (and the .bashrc file).

and I believe these .d directories is quite well-established elegant way.

Being able to extend it from user images is just a small bonus (yes you can export variables with ENV, but you cannot do much more, like put something in  ANDROID_SDK_ROOT or use already exported variable (like ANDROID_SDK_ROOT, ANDROID_HOME, ANDROID_NDK_HOME) which will be available at runtime, but won't be in the extended Dockerfile.

If the main issue is the file being external to Dockerfile, I can create it with here-doc inside Dockerfile, it's just a bit more error prone with Dockerfile syntax getting in a way of simple here-doc.

webbertakken: I don't necessarily think it's a bad way of doing things, but it does increase complexity.

In my opinion my proposed solution of "simply moving the last line of the unity-editor alias file after the android clauses, and make sure that each addition is only added once" is much simpler and more appropriate for the scope of this proposal, which is about supporting multiple modules - not finding the best way to make images extensible.

That said, lets get some additional opinions in here and decide based on what they say.

cc: @davidmfinol @frostebite @GabLeRoux

GabLeRoux: I tend to agree with @webbertakken and his explanation here: https://github.com/game-ci/docker/pull/116#discussion_r641826241

> /etc/profile does this with /etc/profile.d/*sh

Indeed, but as part of the docker image here, I'm not sure I would have had the reflex to look in that folder.

Using a here-doc would be a bit more verbose indeed, but if the goal is to allow user to extend the image and if _moving the last line of that file after the android clauses_ solves that, I would go with @webbertakken's suggestion.

GabLeRoux:
I see that all android checks are failing above, this will most likely need to be fixed before it can be merged 👀 I haven't looked at why it's currently failing tho.

All in all, I think this contribution will be helpful in cases where a single image must be used for multiple platforms. Thanks so much for that contribution! :D

shr-project:
> I see that all android checks are failing above, this will most likely need to be fixed before it can be merged  I haven't looked at why it's currently failing tho.

Is there some successful check since https://github.com/game-ci/docker/pull/97 was merged? The "Validate Android Utils" step fails for me even without any changes from this PR, because it depends on JAVA_HOME from ~/.bashrc while the step passes also -e HOME which will break it unless your HOME is /root (it's '/home/runner' in your CI pipeline).

```
2021-05-29T12:08:56.0092844Z ##[command]/usr/bin/docker run --name a33c1a032c225baa74958bdb83e761f1e8692_15bc0a --label 8a33c1 --workdir /github/workspace --rm -e UNITY_LICENSE -e CHANGESET -e MODULE -e INPUT_IMAGE -e INPUT_RUN -e INPUT_OPTIONS -e INPUT_SHELL -e INPUT_REGISTRY -e INPUT_USERNAME -e INPUT_PASSWORD -e INPUT_DOCKER_NETWORK -e HOME -e GITHUB_JOB -e GITHUB_REF -e GITHUB_SHA -e GITHUB_REPOSITORY -e GITHUB_REPOSITORY_OWNER -e GITHUB_RUN_ID -e GITHUB_RUN_NUMBER -e GITHUB_RETENTION_DAYS -e GITHUB_ACTOR -e GITHUB_WORKFLOW -e GITHUB_HEAD_REF -e GITHUB_BASE_REF -e GITHUB_EVENT_NAME -e GITHUB_SERVER_URL -e GITHUB_API_URL -e GITHUB_GRAPHQL_URL -e GITHUB_WORKSPACE -e GITHUB_ACTION -e GITHUB_EVENT_PATH -e GITHUB_ACTION_REPOSITORY -e GITHUB_ACTION_REF -e GITHUB_PATH -e GITHUB_ENV -e RUNNER_OS -e RUNNER_TOOL_CACHE -e RUNNER_TEMP -e RUNNER_WORKSPACE -e ACTIONS_RUNTIME_URL -e ACTIONS_RUNTIME_TOKEN -e ACTIONS_CACHE_URL -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" -v "/home/runner/work/_temp/_runner_file_commands":"/github/file_commands" -v "/home/runner/work/docker/docker":"/github/workspace" 8a33c1:a032c225baa74958bdb83e761f1e8692
2021-05-29T12:08:57.2402601Z mesg: ttyname failed: Inappropriate ioctl for device
2021-05-29T12:08:57.2470282Z bash: java: command not found
2021-05-29T12:08:57.2601392Z Post job cleanup.
```

It fails the same with docker image built from current main branch:
```
$ docker run -e HOME --rm -it unityci/jansa:editor-android-main /bin/bash -l -c 'java -version'
/bin/bash: java: command not found

$ docker run --rm -it unityci/jansa:editor-android-main /bin/bash -l -c 'java -version'
openjdk version "1.8.0-adoptopenjdk"
OpenJDK Runtime Environment (build 1.8.0-adoptopenjdk-jenkins_2018_05_19_01_00-b00)
OpenJDK 64-Bit Server VM (build 25.71-b00, mixed mode)

$ HOME=/root docker run -e HOME --rm -it unityci/jansa:editor-android-main /bin/bash -l -c 'java -version'
WARNING: Error loading config file: /root/.docker/config.json: open /root/.docker/config.json: permission denied
openjdk version "1.8.0-adoptopenjdk"
OpenJDK Runtime Environment (build 1.8.0-adoptopenjdk-jenkins_2018_05_19_01_00-b00)
OpenJDK 64-Bit Server VM (build 25.71-b00, mixed mode)

$ docker run -e HOME=/root --rm -it unityci/jansa:editor-android-main /bin/bash -c 'java -version'
openjdk version "1.8.0-adoptopenjdk"
OpenJDK Runtime Environment (build 1.8.0-adoptopenjdk-jenkins_2018_05_19_01_00-b00)
OpenJDK 64-Bit Server VM (build 25.71-b00, mixed mode)

$ docker run -e HOME=/foo --rm -it auto-gitlab.lgsvl.net:4567/hdrp/unityci/jansa:editor-android-main /bin/bash -l -c 
'java -version'
```

webbertakken:
Sure, the latest workflows on main were not failing. 

image

The workflows are public and can be found [here](https://github.com/game-ci/docker/actions/workflows/test.yml?query=branch%3Amain)

webbertakken:
The integrity of main looks to be fine. Only 1 iOS build failed because of an activation failure.

editor/unity-editor
```diff
@@ -0,0 +1,11 @@
...
+
+if [ -d /usr/bin/unity-editor.d ] ; then
+    for i in /usr/bin/unity-editor.d/*.sh; do
+      if [ -r $i ]; then
+        . $i
+      fi
+    done
+fi
+
+xvfb-run -ae /dev/stdout "$UNITY_PATH/Editor/Unity" -batchmode "$@"
```
webbertakken, line 11: @shr-project if you could reinstate the previous version without this conf.d loop in a separate file, we'll merge your changes.

shr-project: ok, done

shr-project:
I have no idea why many CI jobs failed now, I see logs in the successful tests, but just "This check failed" for failed like in https://github.com/game-ci/docker/pull/116/checks?check_run_id=3185307822

webbertakken:
> I have no idea why many CI jobs failed now, I see logs in the successful tests, but just "This check failed" for failed like in https://github.com/game-ci/docker/pull/116/checks?check_run_id=3185307822

Likely a GitHub error. Lets rerun them.

webbertakken:
Reported failure here https://github.community/t/many-runners-failing-on-different-stages-seemingly-at-random/193479

DEVELOPMENT.md
```diff
@@ -17,19 +17,19 @@ cd docker
 Build the base image
 
 ```bash
-docker build base -t base
+docker build -t base -f base/Dockerfile .
```
webbertakken, line 20: The readme can be simplified to what it was before. no need to specify the file specifically if you're not passing context. I can revert it in a followup.

PR #132, opened 2021-12-12
Make the comments blocks uniform with 3 spaces

GabLeRoux:
I'm not a huge fan of these big blocks, but I like it when at least everything is aligned ;)


webbertakken:
These blocks can be extremely useful when diffing while moving around blocks of code

PR #135, opened 2021-12-26
Additional Windows Platform Build Targets

AndrewKahr:
#### Changes

- Added support for UWP and tvOS build targets
- Fixed typo in legacy ubuntu editor requested
- Added workflow for building legacy windows editors for UWP and tvOS build targets
- Feature parity with Linux editor dockerfile to install modules after installing the base editor

#### Checklist


webbertakken (approved):
LGTM

webbertakken:
@AndrewKahr shall we add a manual trigger that takes the right variables and merge it so that we can test the workflow?

AndrewKahr:
@webbertakken I think that would be a good idea. How would this look? A separate workflow file with the variables hard coded?

webbertakken:
> @webbertakken I think that would be a good idea. How would this look? A separate workflow file with the variables hard coded?

Great!

Actually you can use `workflow_dispatch` trigger (See [events that trigger workflows](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#manual-events) and [workflow dispatch](https://docs.github.com/en/rest/reference/actions#create-a-workflow-dispatch-event)) in the existing workflows.

We already use `repository_dispatch` which doesn't define the inputs, because that's up to the API call to github. But with `workflow_dispatch` you have to define the inputs.

The output is displayed in part that says `Show hook input` inside our workflow.

AndrewKahr:
@webbertakken Just took a look at workflow_dispatch and I'm not sure how to get it working for our usage. It only allows us to define inputs rather than client_payload which is what we use throughout the workflow file. Could we just manually trigger the repository dispatch with the correct keys set? Or am I missing something?

webbertakken:
> @webbertakken Just took a look at workflow_dispatch and I'm not sure how to get it working for our usage. It only allows us to define inputs rather than client_payload which is what we use throughout the workflow file. Could we just manually trigger the repository dispatch with the correct keys set? Or am I missing something?

Yea we'd need an abstraction for that.

We can have a job that 

- either writes to env
```
echo "MODULE=linux-il2cpp" >> $GITHUB_ENV
```

- or sets output variables:
```
echo "::set-output name=test::hello"
```

based on either the payload or the input, depending on the `event.type`.

**Considered alternatives**

We could also create this trigger logic in the versioning backend (currently it's only able to queue all of them, assuming the workflow works).

The downside of doing it in  backend instead is that we would still need the manual trigger as well, as this will add tremendous value to people forking this repo; as they won't have versioning backend. So it would be useful if forkers are able to run a specific build and publish it.

AndrewKahr:
@webbertakken What if we wrote another workflow to trigger a repository dispatch with the correct parameters using the below action:
https://github.com/marketplace/actions/repository-dispatch

I'm also still a little confused in terms of a manual trigger, if we want to trigger a workflow_dispatch, wouldn't we still need to post to an endpoint to trigger the run? If so, why don't we just trigger a repository dispatch instead with the values set as we need for testing? It wouldn't need to be in the versioning backend, it could be done via a script or another job.

webbertakken:
> @webbertakken What if we wrote another workflow to trigger a repository dispatch with the correct parameters using the below action: https://github.com/marketplace/actions/repository-dispatch
> 
> I'm also still a little confused in terms of a manual trigger, if we want to trigger a workflow_dispatch, wouldn't we still need to post to an endpoint to trigger the run? If so, why don't we just trigger a repository dispatch instead with the values set as we need for testing? It wouldn't need to be in the versioning backend, it could be done via a script or another job.

Right, so in GitHub Actions:

- What you call manual trigger is really called `workflow_dispatch`.
- an automated trigger (trigger manually from a separate system, using an api call) is called `repository_dispatch`.

So if you wanted to make a separate workflow that you would manually trigger, you would have to set up `workflow_dispatch`. If that workflow would send a `payload` to `repository_dispatch`, you would need to get that information from somewhere, when manually triggering it. That's what the inputs are for.

**More practical explanation:**

Or in different words:

If you create a `workflow_dispatch` trigger in a workflow, you get a button to manually trigger it.

Example:

```
on: 
  workflow_dispatch:
    inputs:
      logLevel:
        description: 'Log level'     
        required: true
        default: 'warning'
      tags:
        description: 'Test scenario tags'  
```

Will look like

image

Source: https://github.blog/changelog/2020-07-06-github-actions-manual-triggers-with-workflow_dispatch/

AndrewKahr:
@webbertakken How do we determine whether something was triggered via repository_dispatch or workflow_dispatch? It appears that github.event.event_type is blank for both workflow_dispatch and repository_dispatch. I have also tried github.event.type. I thought perhaps it was a windows issue, but looking at past runs in the docker repo, the type is also blank on ubuntu. Looking at the github.event object, there doesn't appear to be anything pointing to what triggered the workflow to start.

In terms of having a new workflow file, my thinking is we can have one workflow that is for manually triggering both ubuntu and windows (and macOS in the future). The workflow would simply take the inputs and call a repository dispatch on the right event type. This means we don't have to retrofit every single workflow to not use the repo dispatch values. It also avoids the issue of not knowing whether a repository dispatch or workflow dispatch has been triggered. The only code in the workflow would be the inputs and the one call to trigger the repo dispatch thus keeping it straightforward and simple.

webbertakken:
> How do we determine whether something was triggered via repository_dispatch or workflow_dispatch?

If the first input is set then it's `workflow_dispatch` otherwise we can assume `repository_dispatch`.

> In terms of having a new workflow file, my thinking is we can have one workflow that is for manually triggering both ubuntu and windows (and macOS in the future). The workflow would simply take the inputs and call a repository dispatch on the right event type. 

This is essentially what versioning-backend is doing. If we use that approach it would probably be best to add the logic there. 

I believe that idea is the least effective of the two, because there are real downsides:

- we would still have to create an interface to trigger it, which GitHub Actions already provides.
- if we want to provide the manual trigger functionality to people that fork this repo, we'd have to duplicate that logic.

That said, it seems we have different opinions on this one so lets ask input from more peeps.

WDYT @GabLeRoux @davidmfinol @frostebite

AndrewKahr:
@webbertakken Perhaps I am missing the point of the system. My understanding is that we want a debug interface to manually trigger a build to check that it works. I'm not sure what you mean by 
> we would still have to create an interface to trigger it, which GitHub Actions already provides. 
The new workflow would be using workflow_dispatch with the parameters we otherwise would've specified in the editor image workflow thus giving us the same GitHub actions interface to use, though maybe you are referring to another interface that I am missing. I'm also not sure what you mean by  
> if we want to provide the manual trigger functionality to people that fork this repo, we'd have to duplicate that logic.
If they fork the repo, wouldn't it still work the same way? The new workflow wouldn't depend on anything else, it simply takes the workflow dispatch inputs and triggers a repository dispatch. I would think this is actually a benefit because to the editor workflow, both the versioning backend and this debug workflow are triggering it in identical ways, reducing the chance of different behavior between the versioning backend and a debug build. 
Also on forking a repo, wouldn't none of this work regardless because the report to backend steps would fail in the job since people don't otherwise have access to the versioning backend?

AndrewKahr:
For reference, this is what the workflow would look like:
```yaml
name: Debug Image Build
on:
  workflow_dispatch:
      inputs:
        eventType:
          description: 'Dispatch Event Type'
          required: true
          default: 'new_windows_post_2019_2_editor_image_requested'
        repository:
          description: 'Repository Name'
          required: true
          default: 'gameci/docker'
        jobId:
          description: 'Job ID'
          required: true
          default: '0'
        editorVersion:
          description: 'Unity Editor Version'
          required: true
          default: '2020.3.24f1'
        changeSet:
          description: 'Unity Editor Changeset'
          required: true
          default: '79c78de19888'
        repoVersionFull:
          description: 'Docker Repo Version Full'
          required: true
          default: '0.15'
        repoVersionMinor:
          description: 'Docker Repo Version Minor'
          required: true
          default: '15'
        repoVersionMajor:
          description: 'Docker Repo Version Major'
          required: true
          default: '0'
  
jobs:
  dispatch:
    runs-on: ubuntu-latest

    steps:
      - name: Repository Dispatch
        uses: peter-evans/repository-dispatch@v1
        with:
          token: ${{ secrets.REPO_ACCESS_TOKEN }}
          repository: ${{ github.event.inputs.repository }}
          event-type: ${{ github.event.inputs.eventType }}
          client-payload: |
                          '{"jobId": "${{ github.event.inputs.jobId }}", 
                            "editorVersion": "${{ github.event.inputs.editorVersion }}", 
                            "changeSet": "${{ github.event.inputs.changeSet }}", 
                            "repoVersionFull": "${{ github.event.inputs.repoVersionFull }}", 
                            "repoVersionMinor": "${{ github.event.inputs.repoVersionMinor }}", 
                            "repoVersionMajor": "${{ github.event.inputs.repoVersionMajor }}"}'
```

webbertakken:
**Clarification**

> For reference, this is what the workflow would look like:
> ...

I understood this in an earlier post indeed. Referring back to what I responded to that:

> To not overcomplicate things I think it's best to not create yet another workflow, but follow GitHub's reference implementation for manual trigger.

Just to clarify my train of thought, and hopefully clear up possible confusion
In my last post I was weighing these options:

- Triggering single build for debugging from versioning backend
- Triggering debug build by integrating `workflow_dispatch` as a manual trigger for each existing workflow

No longer 

- Triggering a separate workflow that triggers the different existing workflows

So that's where our confusion came from I think.

**Moving forward**

So it seems we're in agreement about using `workflow_dispatch` (so an interface for choosing the inputs is provided by GitHub Actions as shown in the screenshot) and we would still have to choose between:

- Integrating it in each existing workflow (for new images)
- Using a separate workflow that triggers the existing workflows

webbertakken:
Personally I'd still prefer to keep the workflows self-contained and integrate `workflow_dispatch` in each existing workflow.

Upsides:
- No added complexity
- No pollution in workflows, like empty output in the workflow that you're triggering, and switching between different flows when you do
- Good enough for debugging specific workflows as well as publishing specific images

Downsides:
- Some more duplicated code (However, they already share a lot of the logic anyway, and I'm already using diff tool to keep them in sync)
- You have to know whether to use pre-2019.3 or post-2019.3 workflow when triggering one.

For me I think these pros outweight the cons. But it's open for discussion of course! :)

webbertakken (approved):
Amazing work!

PR #139, opened 2022-01-16
Fix Android image size

bkslasher:
Set permissions inside of the builder before copying to reduce resulting image size.

Android image size got increased after implementing Hub 3.0:
[ubuntu-2021.2.6f1-android-0.15.0](https://hub.docker.com/layers/unityci/editor/ubuntu-2021.2.6f1-android-0.15.0/images/sha256-0e4061d36042f35fd224a13a79960ab2b2164ef30d9a2f2e68b0abfc3ec7cb03?context=explore)
[ubuntu-2021.2.6f1-android-0.16.0](https://hub.docker.com/layers/unityci/editor/ubuntu-2021.2.6f1-android-0.16.0/images/sha256-19318e13ce3e21327cb36fc1d51c080a1e7809bc27b14d4c1bec19fb14a75cbc?context=explore)

### DIFF
--- a/images/ubuntu/editor/Dockerfile
+++ b/images/ubuntu/editor/Dockerfile
@@ -20,7 +20,9 @@ RUN for mod in $module; do \
       else \
         unity-hub install-modules --version "$version" --module "$mod" --childModules | tee /var/log/install-module-${mod}.log && grep 'Missing module' /var/log/install-module-${mod}.log | exit $(wc -l); \
       fi \
-    done
+    done \
+	# Set execute permissions for modules
+	&& chmod -R 755 /opt/unity/editors/$version/Editor/Data/PlaybackEngines
 
 ###########################
 #          Editor         #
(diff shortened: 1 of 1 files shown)


webbertakken:
I'm trying to understand this change before merging it.

Could you explain how this works with reducing the image size please?

webbertakken:
See reference above, since we no longer 755 the android folder, the sdk manager is also no longer covered. So should we bring back that line as well?

bkslasher:
> I'm trying to understand this change before merging it. Could you please explain how this reduces the image size?

So, apparently a big recursive chmod increases image size by a lot. See here for example: [ubuntu-2021.2.6f1-android-0.16.0](https://hub.docker.com/layers/unityci/editor/ubuntu-2021.2.6f1-android-0.16.0/images/sha256-19318e13ce3e21327cb36fc1d51c080a1e7809bc27b14d4c1bec19fb14a75cbc?context=explore)
image

This change moves the chmod command into the builder image, so a new layer is not created in the resulting editor image.


> See reference above, since we no longer 755 the android folder, the sdk manager is also no longer covered. So should we bring back that line as well?

`/opt/unity/editors/$version/Editor/Data/PlaybackEngines` is a parent directory for all the modules installed by Unity. That includes SDK, NDK and OpenJDK. I did it like this for simplicity.

webbertakken (approved):
I see. That makes sense indeed. Here's also a [thread](https://stackoverflow.com/questions/30085621/why-does-chown-increase-size-of-docker-image) about it on stack overlfow.

Checking the size of the resulting image [build](https://github.com/game-ci/docker/runs/4831567117?check_suite_focus=true) (2021.2.8f1 android). The build it seems to be 13GB (uncompressed)

image

The previous [build](https://github.com/game-ci/docker/runs/4814466169?check_suite_focus=true) was 18GB (uncompressed)

image

Moving forward with the merge

PR #141, opened 2022-01-22
chore: phase out old workflow tags

webbertakken:
#### Changes

- For base and hub images we build for all base platforms at the same time, tags are shared
- For legacy and specific version editor workflows we keep them separated, so that we can manage workflows for different baseOses separately.

#### Checklist

### DIFF
--- a/.github/workflows/new-ubuntu-base-image-requested.yml
+++ b/.github/workflows/new-ubuntu-base-image-requested.yml
@@ -2,7 +2,9 @@ name: New Ubuntu Base Version ⛭
 
 on:
   repository_dispatch:
-    types: [new_base_image_requested, new_ubuntu_base_image_requested]
+    types:
+      - new_base_images_requested
+      - new_ubuntu_base_image_requested
 
 # Further reading:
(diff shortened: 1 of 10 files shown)


webbertakken:
Found a potential blocker for windows builds:

image

@AndrewKahr these builds ran before right?

AndrewKahr:
I think this might be the first time the windows ones were attempted. I thought they would run with the original PR but it appears that was not the case. I believe it's complaining about using the registry container for the builds. I might be able to just remove it and have it build locally.

AndrewKahr:
@webbertakken Is there a specific reason for using a local registry container vs just building locally on the runner?

webbertakken:
It looks like windows images are not supported [at all](https://github.com/actions/runner/issues/904)?

Hopefully there is some kind of workaround - I think we don't necessarily need GitHub support, just the build to work would go a long way.

AndrewKahr:
Yea that's what I've been seeing. I'm gonna put up a commit that removes reliance on the container registry and just have it build locally on the runner. I think the main purpose of the PR workflow is to make sure it builds so I don't see an issue with this as a workaround.

webbertakken:
Ok. In the meantime I'm going to merge this. Could you reference this PR in your commit please?

PR #143, opened 2022-01-23
chore: one more change to the build triggers

webbertakken:
#### Changes

- Decided to trigger both `ubuntu` and `windows` new editor workflows together. So from now on legacy images are always 2019.2.x or lower, for all base OSes.

Reason is that it reduced complexity in the backend, where we don't have to guarantee multiple workflow dispatches were sent. Goes to idempotence.

#### Checklist

### DIFF
--- a/.github/workflows/new-windows-post-2019-2-editor-image-requested.yml
+++ b/.github/workflows/new-windows-post-2019-2-editor-image-requested.yml
@@ -3,7 +3,8 @@ name: New Windows Editor Version 🗔
 on:
   repository_dispatch:
     types:
-      - new_windows_post_2019_2_editor_image_requested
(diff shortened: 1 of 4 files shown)


.github/workflows/new-windows-post-2019-2-editor-image-requested.yml
```diff
@@ -57,25 +58,25 @@ jobs:
       - name: Setup Build Parameters
         id: buildParameters
         run: |
-              if (${{ github.event.inputs.jobId }})
-              {
-                # Workflow Dispatch
-                echo '::set-output jobId=${{ github.event.inputs.jobId }}'
```
webbertakken, line 63: Everything below here is just autoformatting by prettier

PR #151, opened 2022-02-01
Implement system to handle varying platform module names/compatibility across platforms

AndrewKahr:
This serves as an alternative to #150 and versioning-backend#39. It adds an additional internal github action to handle issues caused by varying module names and compatibility being added for additional target platforms by Unity. It can mark target platforms as incompatible so the system skips building it, for example, GameCI doesn't provide Android on linux below 2019.3 so the action is set such that Android is incompatible on Linux for any version below 2019.3. This also handles modules being renamed over time. For example, uwp-il2cpp became universal-windows-platform in 2019. The action has a mapping defined such that it knows when universal-windows-platform is requested, and the version of unity is below 2019, it will return the correct module name of uwp-il2cpp for the docker build while it remains universal-windows-platform in GameCI labels to reduce complexity in Unity-Builder. Since there is no clean divide between legacy and postX versions, there would need to be work done in versioning-backend every time we needed a new workflow, and more complexity added when windows and linux compatibility diverge. Having this compatibility action should make it easier to handle new modules being added in later versions without needing to make adjustments to the versioning-backend. Unlike #150 and versioning-backend#39, this requires no changes on the versioning-backend for functionality.

#### Changes
- Add new internal GitHub Action `module-compatibility-handler`
- Make windows editor workflows check the `module-compatibility-handler` for module naming and whether a specific module is compatible. (The action is also compatible with Linux, we can add support into the workflows later if deemed necessary)

#### Checklist


.github/workflows/actions/module-compatibility-handler/node_modules/.yarn-integrity
```diff
@@ -0,0 +1,24 @@
+{
```
GabLeRoux, line 1: While reviewing, I asked myself this question:

Do we have to track `node_modules` folders in workflows? I'm not 100% sure of the answer, but we're already doing it here:  
https://github.com/game-ci/docker/tree/main/.github/workflows/actions/report-to-backend

👍

.github/workflows/actions/module-compatibility-handler/index.js
```diff
@@ -0,0 +1,306 @@
...
+  * to allow us to map those renamed modules in a consistent way. The structure is
+  * key: the original name, value: the mapped name.
+  *
+  * Example:
+  * uwp-il2cpp was renamed to universal-windows-platform. We will standardize to
+  * universal-windows-platform so the mapping will be uwp-il2cpp: universal-windows-platform.
+  */
+  const renamedLinuxModules = new TwoWayMap(new Map());
+  const renamedWindowsModules = new TwoWayMap(new Map([['uwp-il2cpp', 'universal-windows-platform']]));
+  const renamedMacModules = new TwoWayMap(new Map());
```
GabLeRoux, line 67: I see some indentation differences, can easily be fixed with linting tools

.github/workflows/actions/module-compatibility-handler/index.js
```diff
@@ -0,0 +1,306 @@
...
+  return moduleSet;
+}
+
+/**
+ * Compares Unity Versions
+ * Returns -1 if versionA is less than versionB
+ * Returns 1 if versionA is greater than versionB
+ * Returns 0 if the versions are identical
+ */
+function compareVersions(versionA, versionB) {
```
GabLeRoux, line 190: I'm wondering if there's an easier way for version comparison. Also I feel like this could break for beta versions if we want to support that for the future (but hey I didn't test this at all)

.github/workflows/actions/module-compatibility-handler/node_modules/.yarn-integrity
webbertakken: We should in fact not do this, even if we are already. It's better to transpile so you can treeshake dependencies.

AndrewKahr: I've converted report-to-backend and module-compatibility-handler to typescript and untracked the node modules. I'm not sure if the configs are set up correctly, I just mimicked what I saw in unity-builder. I also don't know how to set up the precommit hooks with multiple ts projects within a subdirectory of the main repo. If I could get some pointers that would be very helpful.

.github/workflows/actions/module-compatibility-handler/index.js
AndrewKahr: I just added support for looking at the version letter and the number after it. I'm assuming a comes before b which comes before f. Then finally compare by the number after the letter. Let me know if there is anything I am missing there as I am not too well versed with Unity's versioning scheme.

PR #162, opened 2022-03-08
feat: support dedicated server builds

paulpach:
#### Changes

- Add modules for dedicated server build

should fix https://github.com/game-ci/documentation/issues/240
#### Checklist

### DIFF
--- a/images/ubuntu/editor/Dockerfile
+++ b/images/ubuntu/editor/Dockerfile
@@ -28,10 +28,6 @@ RUN echo "$version-$module" | grep -q -v '^2021.[2-3].*linux' \
   && exit 0 \
   || unity-hub install-modules --version "$version" --module "linux-server" --childModules | tee /var/log/install-module-${linux-server}.log && grep 'Missing module' /var/log/install-module-${linux-server}.log | exit $(wc -l);
 
-RUN echo "$version-$module" | grep -q -v '^2021.[2-3].*mac' \
-  && exit 0 \
-  || unity-hub install-modules --version "$version" --module "mac-server" --childModules | tee /var/log/install-module-${mac-server}.log && grep 'Missing module' /var/log/install-module-${mac-server}.log | exit $(wc -l);
-
 RUN echo "$version-$module" | grep -q -v '^2021.[2-3].*windows' \
   && exit 0 \
   || unity-hub install-modules --version "$version" --module "windows-server" --childModules | tee /var/log/install-module-${windows-server}.log && grep 'Missing module' /var/log/install-module-${windows-server}.log | exit $(wc -l);
--- a/images/ubuntu/editor/Dockerfile
+++ b/images/ubuntu/editor/Dockerfile
@@ -24,11 +24,11 @@ RUN for mod in $module; do \
 	# Set execute permissions for modules
 	&& chmod -R 755 /opt/unity/editors/$version/Editor/Data/PlaybackEngines
 
-RUN echo "$version-$module" | grep -q -v '^2021.[2-3].*linux' \
+RUN echo "$version-$module" | grep -q -v '^\(2021.2\|2021.3\|2022\).*\(linux\|base\)' \
   && exit 0 \
   || unity-hub install-modules --version "$version" --module "linux-server" --childModules | tee /var/log/install-module-${linux-server}.log && grep 'Missing module' /var/log/install-module-${linux-server}.log | exit $(wc -l);
 
-RUN echo "$version-$module" | grep -q -v '^2021.[2-3].*windows' \
+RUN echo "$version-$module" | grep -q -v '^\(2021.2\|2021.3\|2022\).*windows' \
   && exit 0 \
   || unity-hub install-modules --version "$version" --module "windows-server" --childModules | tee /var/log/install-module-${windows-server}.log && grep 'Missing module' /var/log/install-module-${windows-server}.log | exit $(wc -l);
 
--- a/images/ubuntu/editor/Dockerfile
+++ b/images/ubuntu/editor/Dockerfile
@@ -24,7 +24,7 @@ RUN for mod in $module; do \
 	# Set execute permissions for modules
 	&& chmod -R 755 /opt/unity/editors/$version/Editor/Data/PlaybackEngines
 
-RUN echo "$version-$module" | grep -q -v '^\(2021.2\|2021.3\|2022\).*\(linux\|base\)' \
+RUN echo "$version-$module" | grep -q -v '^\(2021.2\|2021.3\|2022\).*linux)' \
   && exit 0 \
   || unity-hub install-modules --version "$version" --module "linux-server" --childModules | tee /var/log/install-module-${linux-server}.log && grep 'Missing module' /var/log/install-module-${linux-server}.log | exit $(wc -l);
 
--- a/images/ubuntu/editor/Dockerfile
+++ b/images/ubuntu/editor/Dockerfile
@@ -24,6 +24,18 @@ RUN for mod in $module; do \
 	# Set execute permissions for modules
 	&& chmod -R 755 /opt/unity/editors/$version/Editor/Data/PlaybackEngines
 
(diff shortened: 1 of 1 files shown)


webbertakken (commented):
Hi @paulpach,

This looks good overall.

It would be nice to know how much these additional build target modules add to the images. If it is a lot we might opt to make separate images for these. Hopefully we can go with your solution though.

Could you please verify that the resulting images in fact work as expected when building a project?

paulpach:
Might be worth separating:
Screen Shot 2022-03-08 at 9 16 06 AM

webbertakken:
> Might be worth separating: Screen Shot 2022-03-08 at 9 16 06 AM

Perhaps. Although for Linux and Windows it's quite acceptable it seems.

Do server builds require IL2CPP or not? If yes, then that will not be supported in docker, as MacOS images currently do not exist as per [macoscontainers.org](https://macoscontainers.org/). In that case we can simply support Windows and Linux with docker and MacOS like we already do in unity-builder (https://github.com/game-ci/unity-builder/pull/326)

paulpach:
It does not look like it.  I was able to build a mono dedicated server.

So should I remove the mac support and just leave linux and windows server support?  maybe separate them in 3 prs?  

Honestly I only care about linux servers, I added the other ones for consistency.

webbertakken:
I'm also doubtful about how common the usecase of MacOS server really is.

Perhaps the right course of action here is to  see the actual difference in disk space for Linux and Windows (base target module vs. both base and server target module). If the difference is within 5% we might just go forward with this, without MacOS server support in a first iteration.

What do you think?

webbertakken:
Despite the % increase I'm thinking lets include Linux Server and Windows Server as targets on their respective platforms and leave MacOS Server out of this entirely.

WDYT @GabLeRoux @davidmfinol @frostebite?

davidmfinol:
Agreed with moving forward with just Linux and Windows servers.

I'd also be surprised if anyone ever asked for Mac server.

davidmfinol:
If we do move forward with Linux and Windows servers as targets, would they include IL2CPP?
We've identified that IL2CPP is not necessarily required, as people can do mono server builds.
However, I imagine that some users would want IL2CPP server builds?

images/ubuntu/editor/Dockerfile
```diff
@@ -24,6 +24,18 @@ RUN for mod in $module; do \
...
 	&& chmod -R 755 /opt/unity/editors/$version/Editor/Data/PlaybackEngines
 
+RUN echo "$version-$module" | grep -q -v '^2021.[2-3].*linux' \
+  && exit 0 \
+  || unity-hub install-modules --version "$version" --module "linux-server" --childModules | tee /var/log/install-module-${linux-server}.log && grep 'Missing module' /var/log/install-module-${linux-server}.log | exit $(wc -l);
+
+RUN echo "$version-$module" | grep -q -v '^2021.[2-3].*mac' \
+  && exit 0 \
+  || unity-hub install-modules --version "$version" --module "mac-server" --childModules | tee /var/log/install-module-${mac-server}.log && grep 'Missing module' /var/log/install-module-${mac-server}.log | exit $(wc -l);
+
```
webbertakken, line 34: As discussed, let's remove this entirely

images/ubuntu/editor/Dockerfile
```diff
@@ -24,6 +24,18 @@ RUN for mod in $module; do \
 	# Set execute permissions for modules
 	&& chmod -R 755 /opt/unity/editors/$version/Editor/Data/PlaybackEngines
 
+RUN echo "$version-$module" | grep -q -v '^2021.[2-3].*linux' \
```
webbertakken, line 27: Note that in `unity-builder` linux non-il2cpp builds are built using `base`, not `linux` [currently](https://github.com/game-ci/unity-builder/blob/d975f3b80f2c02df508066ce2a8ac0bc723d4507/src/model/image-tag.ts#L40).

images/ubuntu/editor/Dockerfile
paulpach, line 34:
```suggestion
```

images/ubuntu/editor/Dockerfile
webbertakken: Do we have an example workflow that bootstraps a server build? Does the public api of builder have to change or does it work already?

@davidmfinol do you know?

paulpach: So we need a build with the linux-mono module?

davidmfinol (approved):

images/ubuntu/editor/Dockerfile
webbertakken: Check.

images/ubuntu/editor/Dockerfile
```diff
@@ -24,6 +24,14 @@ RUN for mod in $module; do \
 	# Set execute permissions for modules
 	&& chmod -R 755 /opt/unity/editors/$version/Editor/Data/PlaybackEngines
 
+RUN echo "$version-$module" | grep -q -v '^\(2021.2\|2021.3\|2022\).*\(linux\|base\)' \
```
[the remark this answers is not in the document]
paulpach: so should we add a linux-mono image?  it would contain the linux-mono + linux-server modules.

Separate PR?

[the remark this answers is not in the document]
davidmfinol: We could have a separate linux-mono image, but is the smaller image size worth having additional images? 

`unity-builder` defaults to using the linux-il2cpp image for all Linux builds with version >= 2019.3, and I think it's fine to just use the linux-il2cpp image even for mono builds.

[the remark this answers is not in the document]
paulpach: I see.  ok.

Is there anything else needed to merge this?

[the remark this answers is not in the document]
webbertakken: Yes, there's a parenthesis too many behind `linux` which will not work.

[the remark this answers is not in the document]
webbertakken: > We could have a separate linux-mono image, but is the smaller image size worth having additional images?
> 
> `unity-builder` defaults to using the linux-il2cpp image for all Linux builds with version >= 2019.3, and I think it's fine to just use the linux-il2cpp image even for mono builds.

Exactly, especially for server builds this should be ok. For non-il2cpp non-server builds the base image will the smallest and best option still.

No changes needed on docker side.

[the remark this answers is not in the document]
paulpach: > Yes, there's a parenthesis too many behind `linux` which will not work.

My bad, fixed now

images/ubuntu/editor/Dockerfile
```diff
@@ -24,6 +24,14 @@ RUN for mod in $module; do \
 	# Set execute permissions for modules
 	&& chmod -R 755 /opt/unity/editors/$version/Editor/Data/PlaybackEngines
 
+RUN echo "$version-$module" | grep -q -v '^\(2021.2\|2021.3\|2022\).*linux' \
+  && exit 0 \
+  || unity-hub install-modules --version "$version" --module "linux-server" --childModules | tee /var/log/install-module-${linux-server}.log && grep 'Missing module' /var/log/install-module-${linux-server}.log | exit $(wc -l);
+
+RUN echo "$version-$module" | grep -q -v '^\(2021.2\|2021.3\|2022\).*windows' \
+  && exit 0 \
+  || unity-hub install-modules --version "$version" --module "windows-server" --childModules | tee /var/log/install-module-${windows-server}.log && grep 'Missing module' /var/log/install-module-${windows-server}.log | exit $(wc -l);
```
webbertakken, line 33: Looking at this one last time before merging.

I don't think ${linux-server} and ${windows-server} are defined. Should it be 

`/var/log/install-module-windows-server.log` instead of `/var/log/install-module-${windows-server}.log`?

webbertakken (changes requested):

images/ubuntu/editor/Dockerfile
paulpach: fixed

PR #163, opened 2022-03-13
deps: upgrade libstdc++ to support Autodesk FBX SDK for Unity

jsm174:
#### Changes

`com.autodesk.fbx@4.1` requires `libstdc++ 6.0.28+` (see https://docs.unity3d.com/Packages/com.autodesk.fbx@4.1/manual/index.html)

Unfortunately the version of `libstdc++` that comes with Ubuntu 18 is `6.0.25`. Until the base images are updated to use Ubuntu 20 (which has a much newer version), we need to install an update `libstdc++`.

I was able to make a docker image: 

```
FROM unityci/editor:2021.2.15f1-linux-il2cpp-0
RUN apt-get update && apt-get -y install software-properties-common && add-apt-repository -y ppa:ubuntu-toolchain-r/test && apt-get -y install --only-upgrade libstdc++6
```

This was tested at:

https://github.com/freezy/VisualPinball.Engine/pull/387#issuecomment-1066182772
 
#### Checklist

### DIFF
--- a/images/ubuntu/base/Dockerfile
+++ b/images/ubuntu/base/Dockerfile
@@ -29,7 +29,8 @@ RUN apt-get -q update \
  lsb-release \
  xvfb \
  xz-utils \
- && apt-get clean
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
 
 # Toolbox
 RUN apt-get -q update \
@@ -41,7 +42,8 @@ RUN apt-get -q update \
  openssh-client \
  wget \
  && git lfs install --system --skip-repo \
- && apt-get clean
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
 
 # libstdc++6 upgrade
 RUN apt-get -q update \
@@ -50,7 +52,8 @@ RUN apt-get -q update \
  && apt-get -q install -y --only-upgrade libstdc++6 \
  && add-apt-repository -y --remove ppa:ubuntu-toolchain-r/test \ 
  && apt-get -q remove -y --auto-remove software-properties-common \
- && apt-get clean
+ && apt-get clean \
+ && rm -rf /var/lib/apt/lists/*
 
 # Disable default sound card, which removes ALSA warnings
 RUN /bin/echo -e 'pcm.!default {\n\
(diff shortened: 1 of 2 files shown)


webbertakken:
We usually only add target-platform-specific additions to  `editor` image. These changes should be made in the base image instead (`images/ubuntu/base/Dockerfile`).

Could you please give an indication of how much space is used by the base image before and after?

jsm174:
> We usually only add target-platform-specific additions to `editor` image. These changes should be made in the base image instead (`images/ubuntu/base/Dockerfile`).
> 
> Could you please give an indication of how much space is used by the base image before and after?

Thanks. I was going to do base and then I wasn't sure. Reverted the change, and updated accordingly. 

As for size, it looks like it's 25mb.

```
docker images -a unityci/base:ubuntu-latest
REPOSITORY     TAG             IMAGE ID       CREATED       SIZE
unityci/base   ubuntu-latest   841ddfea6757   2 weeks ago   474MB

docker images -a jsm174/unityci_base:ubuntu-latest
REPOSITORY            TAG             IMAGE ID       CREATED          SIZE
jsm174/unityci_base   ubuntu-latest   069fbd6ea096   41 minutes ago   499MB
```

I feel like more could be cleaned up, but I must be missing something. I added `software-properties-common` just to get `add-apt-repository` and then I remove it after installation.

jsm174:
@webbertakken - I was looking at the editor docker file, and noticed all  

`apt-get clean` are followed by `rm -rf /var/lib/apt/lists/*`

If we do that here,

```
docker images -a jsm174/unityci_base:ubuntu-latest
REPOSITORY            TAG             IMAGE ID       CREATED         SIZE
jsm174/unityci_base   ubuntu-latest   c1e5176d1603   6 seconds ago   479MB
```

The change is ~4MB.  We could that for the first block as well and reduce the size of the image as well?

jsm174:
Screen Shot 2022-03-14 at 8 59 35 AM

jsm174:
Sure. Ahh, you were right, only a decrease of about 27mb. 

Screen Shot 2022-03-17 at 3 06 17 PM

webbertakken:
**Context**

Tried the changes locally to make a more exact comparison of `base` image.

- `before` is the current version on `main`.
- `without-lists` same as before but with the apt-get lists removed in first 2 `RUN` instructions
- `without-lists-upgrade-lib` same as before but with the lib upgraded

**Proof**

image

**Conclusion**

Upgrading the lib in this MR increases size by 4 megs, which is acceptable, as it allows (experimental) unity packages to work out of the box.

PR #197, opened 2023-01-27
Add zip and unzip to the base image

AndrewKahr:
#### Changes

- Add zip and unzip to the base image (1205 kB) as some Unity packages leverage it for working with Android libraries. 

One example is Microsoft App Center for Unity as it uses it to unzip aar's for Android builds to modify settings within the aar before rezipping it for the build process. The editor install docker image also utilizes unzip for certain Android setup steps. 

#### Checklist

### DIFF
--- a/images/windows/hub/Dockerfile
+++ b/images/windows/hub/Dockerfile
@@ -1,7 +1,6 @@
 ARG baseImage="unityci/base:windows-latest"
 
 FROM $baseImage
-ARG hubVersion="3.0.0"
 
 # Install unity hub
-RUN choco install unity-hub --version=%hubVersion --no-progress -y
+RUN choco install unity-hub --no-progress -y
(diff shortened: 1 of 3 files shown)


images/windows/hub/Dockerfile
```diff
@@ -1,7 +1,6 @@
 ARG baseImage="unityci/base:windows-latest"
 
 FROM $baseImage
-ARG hubVersion="3.0.0"
 
 # Install unity hub
-RUN choco install unity-hub --version=%hubVersion --no-progress -y
```
webbertakken: Removing the version lock makes the build non-deterministic. This can be problematic, because any possible errors from version bumps should be done in development time, not during publishing time.

AndrewKahr: Oh I moved that to my other PR #198 since its a bit more in line with the PR title. Should I move it back to this PR?

webbertakken: My suggestion is to always lock the version, and keep it as a hardcoded value somewhere.

AndrewKahr: I'll move it back then

webbertakken (approved):
Some references to docs or example libraries would be useful to have in the rationale.

Other than that LGTM.

PR #198, opened 2023-01-27
Fix deprecations and errors

AndrewKahr:
#### Changes

- Replace setoutput with the new environment output method
- Update node12 to node16
- Add the test package to unity modules to fix build errors on 2022.2 (Only applied to Unity 2020+ as that is when the package came into support)
- Lock UnityHub version to a defined value rather than the latest version at the time of build (matches Linux behavior)
- Add --no-sandbox flag to UnityHub in preparation for the latest v3 release that should allow us to build 2022.2 properly

#### Checklist


AndrewKahr:
I can't tell why but I've added --no-sandbox to the editor dockerfile when installing editors via the hub but I don't see the pr CI running with that flag. Does it not test PR code?

davidmfinol (approved):
LGTM

PR #206, opened 2023-04-11
Misc Improvements and Bug Fixes

AndrewKahr:
#### Changes

- Add tests for Windows images
- Consolidate testing workflows
- Fix checking out main instead of latest tag for building images
- Fix using GITHUB_ENV rather than GITHUB_OUTPUT which broke Windows image builds
- Improve error checking on Windows editor installations
- General improvements to error checking on editor/module installations
- Hide node error spam from Unity Hub by piping stderr to null. This is ok because the errors we care about end up on stdout

#### Checklist


images/ubuntu/hub/Dockerfile
```diff
@@ -16,8 +16,11 @@ RUN sh -c 'echo "deb https://hub.unity3d.com/linux/repos/deb stable main" > /etc
  && apt-get -q install -y "unityhub=$hubVersion" \
  && apt-get clean
 
+# Bug in xvfb-run that causes it to redirect stderr to stdout. We want it separate
```
GabLeRoux, line 19: Do you have details about that bug?

AndrewKahr: https://bugs.launchpad.net/ubuntu/+source/xorg-server/+bug/1059947
https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=868876

GabLeRoux (commented):
I think it would be worth it at some point to extract some scripts from some of the workflow files for readability (ex; in `.github/workflows/test.yml`). Other than that, this looks good to me 👍 🤯

GabLeRoux (approved):

GabLeRoux:
The following two jobs both failed with a timeout, but everything else is green.

https://github.com/game-ci/docker/actions/runs/4675713024/jobs/8286515438?pr=206
https://github.com/game-ci/docker/actions/runs/4675713024/jobs/8286515018?pr=206

Are we ok with that?

AndrewKahr:
I checked the timeouts and they are from Unity not exiting despite completing the build successfully in well under 15 minutes. This is a known issue but out of scope for this PR so it is acceptable. The images themselves are building and installing the correct modules which is the main thing that matters for the scope of this PR.

PR #210, opened 2023-04-17
Disable Legacy Windows Image Builds

AndrewKahr:
#### Changes

- New error checking has revealed that legacy Windows images weren't actually working so they will be removed to unblock the build pipeline

#### Checklist


webbertakken:
The code that triggers this repository_dispatch needs to be updated first.

AndrewKahr:
@webbertakken Not sure I follow. The trigger is a general event of `new_legacy_image_requested`. I was under the impression the backend will fire this event and when the builds start they report to the backend which saves the editor version and module combinations. Is there somewhere in the backend that knows the version/module pairings too?

webbertakken:
Right. `new_legacy_editor_image_requested` is fired from versioning backend.
But now I see what you mean (I didn't understand it from the description at first).  You just delete the workflow for the Windows side. LGTM 👍🏻


Have you also considered only removing the trigger, so that people that fork this repo can still use the workflow using the manual build workflow?
Will leave it up to you whether that's better or not.

webbertakken (approved):

AndrewKahr:
Great idea! I've gone ahead and just removed the dispatch trigger.

Issue #213, opened 2023-06-16
Support for Windows Server 2022


mschultz-aofl:
**Context**

Windows Server 2019 (1809) will be EOL in about 6 months. Attempting to run a Windows-based Editor container in Windows Server 2022 returns 
`ERROR: Job failed (system failure): Error response from daemon: hcs::CreateComputeSystem 495d65b4034d0510367c8c9daa8a60ff46591a2653d5aebe4e6921620e3f0e40: The container operating system does not match the host operating system. (exec.go:78:0s)`

**Suggested solution**

Add support for Windows server 2022 in the Base template - It's unclear if we need to support both going forward, or if we should simply stop supporting Win2019 as MS is doing

**Considered alternatives**

Not really one, unless we want to keep using an EOL OS for il2cpp builds

**Additional details**

https://learn.microsoft.com/en-us/lifecycle/products/windows-server-2019

AndrewKahr:
Images are now Server 2022 in v3

Issue #209, opened 2023-04-16
Missing WebGL building docker container for Windows


SimonDarksideJ:
**Bug description**

We recently switched to a self-hosted agent to reduce costs, however, we are no longer able to Build for the WebGL platform as there is no Windows based WebGL container.

Checking on dockerhub and the only WebGL Editor containers are for Ubuntu, which when building through WSL/Docker on Windows, the Builder action looks for a Windows variant of the container which does not exist.

**How to reproduce**

- Try to create a WebGL build on Windows

**Expected behavior**

Should still be able to build WebGL on windows

AndrewKahr:
Images for Android on Windows now exist for v3. We don't have webgl for windows as Ubuntu builds those already and Windows doesn't provide additional functionality (Android on Windows has Burst, Linux does not).

Issue #180, opened 2022-05-17
Burst Android NDK problem 


jjaine:
**Bug description**

I'm using the `ubuntu-2021.3.2f1-android-1.0.1` image to build the project. However the build fails with the message 
```
Burst requires the android NDK to be correctly installed (it can be installed via the unity installer add component) in order to build a standalone player for Android with ARMV7A_NEON32
The environment variable ANDROID_NDK_ROOT is undefined or empty, is the Android NDK correctly installed?
```

This problem arised when I updated from `ubuntu-2020.3.30f1-android-0.15.0` to `ubuntu-2021.3.2f1-android-1.0.1`. If I disable Burst compilation for Android, then the build seems to work.

**How to reproduce**
- Building project with batch mode `unity-editor -quit -batchmode -executeMethod Build.AndroidBuild -apkName "$APK_NAME" -logFile -` on Bitbucket Pipelines.

**Expected behavior**
- Build succeeds

**Additional details**
The packages I have installed are
```
Registered 54 packages:
  Packages from [https://packages.unity.com]:
    com.unity.ide.rider@3.0.13 (location: /playerconsole/Library/PackageCache/com.unity.ide.rider@3.0.13)
    com.unity.ide.visualstudio@2.0.14 (location: /playerconsole/Library/PackageCache/com.unity.ide.visualstudio@2.0.14)
    com.unity.ide.vscode@1.2.5 (location: /playerconsole/Library/PackageCache/com.unity.ide.vscode@1.2.5)
    com.unity.localization@1.0.5 (location: /playerconsole/Library/PackageCache/com.unity.localization@1.0.5)
    com.unity.mobile.android-logcat@1.2.3 (location: /playerconsole/Library/PackageCache/com.unity.mobile.android-logcat@1.2.3)
    com.unity.timeline@1.6.4 (location: /playerconsole/Library/PackageCache/com.unity.timeline@1.6.4)
    com.unity.mathematics@1.2.6 (location: /playerconsole/Library/PackageCache/com.unity.mathematics@1.2.6)
    com.unity.burst@1.6.5 (location: /playerconsole/Library/PackageCache/com.unity.burst@1.6.5)
    com.unity.addressables@1.19.19 (location: /playerconsole/Library/PackageCache/com.unity.addressables@1.19.19)
    com.unity.nuget.newtonsoft-json@3.0.2 (location: /playerconsole/Library/PackageCache/com.unity.nuget.newtonsoft-json@3.0.2)
[...]

PR #231, opened 2023-10-28
Update Windows to 2022 and enable Android Windows Images

AndrewKahr:
#### Changes

- Update windows images to 2022 from 2019. We need to do this because 2019 will go EOL in January which means Github will start failing builds using it and ultimately remove it within a few months after EOL based on how they operated with Server 2016. This means people will need to upgrade their workflows to use windows-2022 instead of windows-2019 and update any self hosted runners to support server-2022 containers (either use Windows Server 2022 or Windows 11 as their host)
- Setup Android dependencies to allow for Android builds on Windows. This enables using Burst on Android builds which is a major performance booster.
- Update several action workflow versions to fix dependabot warnings
- Setup better test exclusion patterns for Windows
- Move Unity install path to C:/UnityEditor. This fixes old versions of sdkmanager that don't handle spaces in the path. Using the UNITY_PATH environment variable will mitigate any changes like this in the future in downstream scripts.

Note builds will fail until unity-builder gets merged with the windows related changes. [Here](https://github.com/game-ci/docker/actions/runs/6664917005) is a representative run of the test images.

#### Related PRs
* game-ci/unity-builder#588 (This needs to be merged first)
* game-ci/documentation#433 (This needs to be merged when changes go live)

There will likely be one more PR for bumping the major version of the docker version number in Unity Builder

#### Related Issues
* #180 
* #213 
* #209 

#### Checklist


.github/workflows/new-windows-post-2019-2-editor-image-requested.yml
```diff
@@ -7,25 +7,25 @@ on:
...
+        description: "Job ID"
         required: true
-        default: 'dryRun'
+        default: "dryRun"
       editorVersion:
-        description: 'Unity Editor Version'
+        description: "Unity Editor Version"
         required: true
-        default: '2020.3.24f1'
+        default: "2020.3.24f1"
```
webbertakken, line 16: Last time I checked yaml usually uses single quotes. This includes all GHA examples. 

Unless this is done by our linter rules I suggest we also stick to single quotes instead.

webbertakken (approved):

.github/workflows/new-windows-post-2019-2-editor-image-requested.yml
AndrewKahr: This seems to have been done by the GitHub actions vscode plugin as when I saved in vscode it auto replaced them all. Since it's by GitHub themselves I figured it would be fine 🤷‍♂️.

davidmfinol (approved):

davidmfinol (approved):

PR #284, opened 2026-05-01
refactor: modernize report-to-backend action

frostebite:
## Summary

- **Replaces abandoned `httpie` package** (last release May 2019) with native `fetch()`, removing deprecated `url.parse()` calls that will break in future Node.js versions
- **Adds retry with exponential backoff** for transient 5xx errors (3 attempts at 2s/4s/8s intervals) — this directly fixes the root cause of the false failures seen on [Apr 25 run #24942210259](https://github.com/game-ci/docker/actions/runs/24942210259) where transient backend 500s caused successfully-published images to be incorrectly reported as "failed"
- **Bundles with `@vercel/ncc`** into a single `dist/index.js`, removing ~300 committed `node_modules` files from the repository
- **Upgrades from `node20` to `node24`**, ahead of the [June 2 2026 deadline](https://github.blog/changelog/2025-09-19-deprecation-of-node-20-on-github-actions-runners/) when GitHub will force Node.js 24 for all actions

## What changed

| File | Change |
|------|--------|
| `index.js` | Replace `httpie.post()` with `fetch()` + `postWithRetry()` wrapper |
| `action.yml` | `using: node20` → `node24`, `main: index.js` → `dist/index.js` |
| `package.json` | Remove `httpie`, add `@vercel/ncc` devDependency, add `build` script |
| `dist/index.js` | New bundled output (single file, no external deps at runtime) |
| `node_modules/` | **Deleted** (~300 files) — now in `.gitignore` |
| `yarn.lock` | **Deleted** — replaced by `package-lock.json` |
| `.gitignore` | Added `node_modules/` path for the action |

## Retry behavior

The new `postWithRetry()` function:
- Retries on HTTP 5xx responses up to 3 times with exponential backoff (2s, 4s, 8s)
- Logs warnings via `core.warning()` on each retry so the behavior is visible in workflow logs
- Does **not** retry on 409 (duplicate dispatch) — these are handled gracefully as before
- Does **not** retry on 4xx client errors — these fail immediately

## Test plan

- [ ] Trigger a dry run build and verify all three report statuses (started/published/failed) work correctly
- [ ] Verify the bundled `dist/index.js` runs correctly under Node.js 24
- [ ] Confirm retry warnings appear in logs when backend returns transient errors

🤖 Generated with [Claude Code](https://claude.com/claude-code)


webbertakken (approved):
Excellent. I never even realised we had a node modules in there

PR #287, opened 2026-05-07
fix: skip duplicate server module installs

frostebite:
## Summary
- skip auto-installing linux-server/windows-server when Unity Hub already selected them via child modules
- prevents Unity Hub failures like 'module windows-server is already selected in modules.json'

## Verification
- not run: Docker template-only change


images/windows/hub/Dockerfile
```diff
@@ -1,6 +1,12 @@
+# escape=`
```
webbertakken, line 1: Ehhhh, I don't think the reasoning behind this is correct

webbertakken (approved):

